PR #3, opened 2022-04-25
Use snake_case for config file

duncanjbrown:
This doesn't make any difference in practice, but it will save new clients the trouble of renaming this file to appease Rubocop


misaka (approved):
🐍 FTRW!!!

PR #10, opened 2022-05-20
Make Rails a development_dependency

duncanjbrown:
It's not a direct dependency — it's only necessary for executing the code in an environment where Rails isn't present.

**Depends on #9**


misaka (approved):
🚀

PR #11, opened 2022-05-26
add have_sent_analytics_event rspec matcher 

misaka:
This matcher gets (optionally) installed as part of the generator.


misaka:
As discussed with @duncanjbrown, we're going to try to change this so that instead of pulling this into the generator as a template, the app that wants to use this matcher will have to require it into the spec that uses the matcher.

lib/dfe/analytics/rspec/matchers/have_sent_analytics_event_types.rb
```diff
@@ -0,0 +1,32 @@
...
+    original_enqueued_jobs_count = queue_adapter.enqueued_jobs.count
+    actual.call
+    in_block_jobs = queue_adapter.enqueued_jobs.drop(original_enqueued_jobs_count)
+
+    @enqueued_event_types = jobs_to_event_types(in_block_jobs)
+
+    event_types.each do |event_type|
+      expect(@enqueued_event_types).to include(event_type.to_s)
+    end
+    # expect(event_types.map(&:to_s) & @enqueued_event_types).not_to be_blank
```
duncanjbrown, line 22: Stray comment

duncanjbrown, line 22: Stray comment

misaka:
[Roll out BigQuery client code across BAT services (dfe-analytics)](https://trello.com/c/neK2rSqu/563-roll-out-bigquery-client-code-across-bat-services-dfe-analytics)

lib/dfe/analytics/rspec/matchers/helpers.rb
```diff
@@ -0,0 +1,29 @@
...
+          def jobs_to_event_types(jobs)
+            jobs.map do |j|
+              j[:args].first.map do |e|
+                e.fetch('event_type')
+              end
+            rescue StandardError
+              # Parsing the job args above makes a couple of assumptions that may
+              # raise an error. Treat these as non-analytics events by returning
+              # 'nil' for the type.
+              nil
```
duncanjbrown, line 22: Should we log here, or are all errors definitely to be ignored?

duncanjbrown, line 22: Should we log here, or are all errors definitely to be ignored?

spec/dfe/analytics/rspec/matchers/have_been_enqueued_event_types_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    ActiveJob::Base.queue_adapter = :test
+  end
+
+  let(:web_request_event) { DfE::Analytics::Event.new.with_type('web_request') }
+  let(:update_entity_event) { DfE::Analytics::Event.new.with_type('update_entity') }
+
+  it 'passes when the given event type was triggered' do
+    DfE::Analytics::SendEvents.do([web_request_event.as_json])
+
+    expect(:web_request).to have_been_enqueued_event_types
```
duncanjbrown, line 22: This reads a bit oddly — could it be

```ruby
expect(:web_request).to have_been_enqueued_for_analytics
```

or sth?

duncanjbrown, line 22: This reads a bit oddly — could it be

```ruby
expect_analytics_event_to_have_been_enqueued(:web_request)
```

or even 

```ruby
expect_analytics_web_request_event_to_have_been_enqueued # etc...
```

duncanjbrown (commented):
Looking vg — nits only inline incl one about naming — wdyt?

Should `RSpec` be nested under `Testing`? `DfE::Analytics::Testing::RSpec` is verbose, but definitely not confusing.

What do you think of including matchers in our own test suite without the `require_relative`?

Screenshot 2022-06-08 at 10 47 40

tomas-stefano (approved):

misaka:
> Should `RSpec` be nested under `Testing`? `DfE::Analytics::Testing::RSpec` is verbose, but definitely not confusing.

Ah, I didn't see `dfe/analytics/testing`. Is the intention for these to be used in using projects? Or only in `dfe-analytics`? If the former then ya, let's keep them together, but also shouldn't we document those helpers? If the latter, then I wonder why those are in `lib` and not `spec`?

> What do you think of including matchers in our own test suite without the `require_relative`?

💯  good call!

lib/dfe/analytics/rspec/matchers/helpers.rb
misaka: I think they can safely be ignored. I can't see a way to be tripped up by this when using the matcher in your spec. The only way to trip up that comes to mind is that if you're changing the matcher code, you might be a little bit surprised if you have an error in there, but I think that's rare and that if it happens, it won't be hard to debug when you realise that all the `event_types` returned are `nil`.

Do you think that's good enough justification to swallow errors here?

misaka: duplicate of the above

lib/dfe/analytics/rspec/matchers/have_sent_analytics_event_types.rb
misaka: duplicate of the above

spec/dfe/analytics/rspec/matchers/have_been_enqueued_event_types_spec.rb
misaka: I agree that this reads oddly, I was justifying doing it this way by looking at existing RSpec matchers [`have_been_enqueued`](https://relishapp.com/rspec/rspec-rails/docs/matchers/have-been-enqueued-matcher). It seemed to me that this would provide for the least surprise for users ... but then, maybe you're demonstrating that that's not true? What do you think in light of the existing RSpec matchers?

re: `expect_analytics_web_request_event_to_have_been_enqueued` I did play with having `have_sent_request_analytics_event_types` and `have_sent_entities_analytics_event_types`, but that added complexity at a different levels and I just removed them. Rubocop was whinging about having them in the same file as the `have_sent_analytics_event_types`, and they wouldn't be usable in feature tests such as the [ones I was writing for Publish](https://github.com/DFE-Digital/teacher-training-api/pull/2662/files#diff-34d199d14ec3f3114d2943dc554c19749a671a48ff922d41ac54f61108a35657R47-R49) which would expect both `web_requests` and one of the entity events. 

```ruby
  def then_it_has_sent_analytics_events
    expect([:web_request, :create_entity]).to have_been_enqueued_event_types
  end
```

However, as we've both thought of it, maybe there's something to it. To make this work, I'd think we'd want:

- `expect_analytics_web_request_event_to_have_been_enqueued`
- `expect_analytics_entity_event_to_have_been_enqueued`
- `expect_any_analytics_event_to_have_been_enqueued`

then perhaps for completeness:

- `have_sent_web_request_analytics_events`
- `have_sent_entity_analytics_events`
- `have_sent_any_analytics_events`

and then maybe we could put these into `lib/dfe/analytics/rspec/matchers', or another helper file, to keep Rubocop from whinging.

Or we could just add documentation for the available `event_types` and call it a day. 😉  What do you think?

PR #16, opened 2022-06-10
CHANGELOG.md and release process

duncanjbrown:


defong (approved):

misaka:
Annoyingly, `github_changelog_generator` doesn't install on my system:

```
$ be gem install github_changelog_generator
ERROR:  Error installing github_changelog_generator:
	There are no versions of io-event (~> 1.0.0) compatible with your Ruby & RubyGems. Maybe try installing an older version of the gem you're looking for?
	io-event requires Ruby version >= 3.0. The current ruby version is 2.7.2.137.
```

And I can't install ruby 3.1.x yet for ... reasons (OpenSSL issues ... 😱 )

We could add a `.tool-versions` to ensure Ruby 3.x is used and add `github_changelog_generator` and `gem_release` to the dev dependencies, however ... that seems a bit draconian, and would get in the way of testing on different versions of Ruby, if we ever wanted to. 🤦

misaka (approved):
Despite the issues I encountered, I say we get this in. Better to have a way forward than not.

PR #17, opened 2022-06-17
make current_user optional

misaka:
For controllers that serve public pages/apis, it doesn't make sense to require a 'current_user' method.

This PR allows the `current_user` method to be missing, and adds a test for that. It also adds a test for `current_namespace` to be optional.

### DIFF
--- a/lib/dfe/analytics/requests.rb
+++ b/lib/dfe/analytics/requests.rb
@@ -16,7 +16,7 @@ module DfE
                                              .with_type('web_request')
                                              .with_request_details(request)
                                              .with_response_details(response)
-                                             .with_user_and_namespace(current_user, try(:current_namespace))
+                                             .with_user_and_namespace(try(:current_user), try(:current_namespace))
                                              .with_request_uuid(RequestLocals.fetch(:dfe_analytics_request_id) { nil })
 
         DfE::Analytics::SendEvents.do([request_event.as_json])
--- a/spec/dfe/analytics/requests_spec.rb
+++ b/spec/dfe/analytics/requests_spec.rb
@@ -5,13 +5,18 @@ RSpec.describe DfE::Analytics::Requests, type: :request do
     controller = Class.new(ApplicationController) do
       include DfE::Analytics::Requests
     end
+    public_api_controller = Class.new(PublicApiController) do
+      include DfE::Analytics::Requests
+    end
 
     stub_const('TestController', controller)
+    stub_const('TestPublicApiController', public_api_controller)
   end
 
   around do |ex|
     Rails.application.routes.draw do
       get '/example/path' => 'test#index'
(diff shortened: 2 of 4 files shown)


tomas-stefano (approved):

lib/dfe/analytics/requests.rb
```diff
@@ -16,7 +16,7 @@ def trigger_request_event
                                              .with_type('web_request')
                                              .with_request_details(request)
                                              .with_response_details(response)
-                                             .with_user_and_namespace(current_user, try(:current_namespace))
+                                             .with_user_and_namespace(try(:current_user), try(:current_namespace))
```
defong, line 19: probably nicer to go symmetry...

```suggestion
                                             .with_user(try(:current_user))
                                             .with_namespace(try(:current_namespace))
```

duncanjbrown: I agree

spec/dfe/analytics/requests_spec.rb
```diff
@@ -5,13 +5,18 @@
     controller = Class.new(ApplicationController) do
       include DfE::Analytics::Requests
     end
+    public_api_controller = Class.new(PublicApiController) do
+      include DfE::Analytics::Requests
+    end
 
     stub_const('TestController', controller)
+    stub_const('TestPublicApiController', public_api_controller)
```
duncanjbrown, line 13: Is this name pulling in a bit too much context from Publish? the important thing is that there is no user, so how about something like `TestUnauthenticatedController`?

misaka:
> 👍 I agree with this decision! Please could you update the README too?

I updated the comment around `current_user`, did you want me to add some more documentation?

spec/dfe/analytics/requests_spec.rb
misaka: Sure. I'll rename.

lib/dfe/analytics/requests.rb
misaka: Ok, I'll split out these into two methods.

duncanjbrown:
> > 👍 I agree with this decision! Please could you update the README too?
> 
> I updated the comment around `current_user`, did you want me to add some more documentation?

oh, so you did — sorry! LGTM!

lib/dfe/analytics/requests.rb
misaka: I've done this, but also changed the code to be slightly more descriptive:

```ruby
        request_event.with_user(current_user)           if respond_to? :current_user
        request_event.with_namespace(current_namespace) if respond_to? :current_namespace
```

One effect this has is that if these methods aren't implemented then `user_id` and `namespace` won't be part of the payload to BQ. I don't think this will change any behaviour in BQ.

duncanjbrown (approved):

PR #18, opened 2022-06-21
Add release task

misaka:
Some minor automation around the release process, this would allow us to have the version bump and changelog updated in a single commit, and with more automation.

- Should we automate further, and include the `git commit` and `git tag` as part of the Rake task? Leaving the `git push` for the user to do once they've examined the change.

- Is the `v` in `v1.0.0` important? This change does away with it.


Rakefile
```diff
@@ -3,3 +3,25 @@
...
+  bump_version = args.fetch(:version)
+
+  current_branch = `git branch --show-current`.chomp
+  raise 'could not get current branch' if current_branch.empty?
+
+  sh 'git', 'checkout', '-b', 'new-release' if current_branch == 'main'
+
+  sh 'gem', 'bump', '-v', bump_version, '--no-commit'
+
+  version = `ruby -rrubygems -e 'puts Gem::Specification::load("dfe-analytics.gemspec").version'`.chomp
```
duncanjbrown, line 19: How about

```suggestion
  version = `bundle exec ruby -e 'puts DfE::Analytics::VERSION'`.chomp
```

Rakefile
```diff
@@ -3,3 +3,25 @@
...
+  raise 'could not retrieve version' if version.empty?
+
+  sh 'github_changelog_generator', '--no-verbose', '--future-release', version
+
+  sh 'git', 'commit', '-a', '-m', version
+
+  sh 'git', 'tag', version
```
duncanjbrown, line 26: `gem tag` will do the same without requiring `version` and will include the `v...` (that's why it was there before!)

Rakefile
duncanjbrown: Ah—see below!

Rakefile
```diff
@@ -3,3 +3,25 @@
 require 'bundler/setup'
 
 require 'bundler/gem_tasks'
+require 'debug'
```
duncanjbrown, line 6: Do we need this here?

duncanjbrown (commented):
🙌 This is a great improvement! I think `gem-release` can further simplify for us?

Rakefile
misaka: Thought about that, but liked the agnosticism of using `Gem::Specification`. Happy either way though, happy to switch this.

Rakefile
misaka: Ah, missed that. I'll have another look.

Rakefile
misaka: Nope, this is just left-overs.

Rakefile
misaka: Changed this to use `gem tag` and 'v' + version

Rakefile
misaka: Changed.

Rakefile
```diff
@@ -3,3 +3,30 @@
...
 require 'bundler/gem_tasks'
+
+desc 'Prepare a new version for release, version can be major, minor, patch or x.y.z (as per gem-release gem)'
+task :prepare_release, %i[version] do |_, args|
+  bump_version = args.fetch(:version)
+
+  current_branch = `git branch --show-current`.chomp
+  raise 'could not get current branch' if current_branch.empty?
+
+  sh 'git', 'checkout', '-b', 'new-release' if current_branch == 'main'
```
duncanjbrown, line 14: Will this fail the second time unless we delete the branch locally? Maybe we should name the branch after the version?

misaka: Ya, so that's tricky because we don't know the version until the next command ... we could use `gem bump --pretend` and parse the output, I considered that, but went with simple.

wrt failing the second time, the default when we merge a PR is to delete the branch, so I assumed this would actually not be an issue in practice. Of course, that risks completely backfiring ... wdyt? Worth the extra complexity of trying to extract the new version number?

misaka: Oh, the other option, of course, is bump as we do, without committing, and then creating the branch after. I was worried that creating the branch might fail, for reasons unknown, and we'd be straight back to messy-land.

duncanjbrown (approved):

PR #20, opened 2022-06-23
Add .tool-versions

duncanjbrown:
Because the `async` gem, which is a (distant) dependency of `github_changelog_generator`, introduces a Ruby version constraint of `>=3.1.1` for the `2.x` series, bundling on a machine with Ruby `3.1.1` installed will result in a different `Gemfile.lock` (ie with `async > 2`) than bundling on a machine with an earlier Ruby version (which gets `async < 2`). If you're in the second group and your colleague is in the first, `bundle install` fails with an unhelpful error:

```
Your bundle is locked to async (2.0.3) from rubygems repository https://rubygems.org/ or installed locally, but that version can no longer be found in that source. That means the
author of async (2.0.3) has removed it. You'll need to update your bundle to a version other than async (2.0.3) that hasn't been removed in order to install.
```

In fact the behaviour of bundler here is quite confusing because it says that `async (2.0.3)` has been *removed* by the author, which isn't actually true — it's just that the version in the `Gemfile.lock` can't be reconciled with the local Ruby version. This incorrect error is a bug in `bundler`! They've fixed it, but haven't released it yet, afaics.

https://github.com/rubygems/rubygems/pull/5525

Note that this only affects dev environments: `github_changelog_generator` is a development dependency only so the official Ruby version supported by `dfe-analytics` is still `2.7`.

Alternative to pinning a Ruby version for dev is to explicitly depend on `async` `< 2.0` in the `gemspec`, but that's a bit weird and mysterious considering it's not our direct dependency.


misaka:
Ya, I was thinking of the same. Would you be ok with using `asdf`, though? I think it's a better solution and more ... future proof.

duncanjbrown:
Good shout. Amended!

misaka (approved):
🎸

PR #21, opened 2022-06-24
Simplify backfill process

duncanjbrown:
**Depends on #20.**

### The old way 🪨 

When bulk-importing existing records into BQ, we used to:

- get `$batch_size` records from the table using `find_in_batches`
- loop over them 
  - enqueue one `SendEvents` job per record
- before the next batch, sleep for `$sleep_time` so we don't enqueue too many (??)

This wasn't great because we ran this via rake over ssh, and often the loop would outlive the ssh session timeout. Or the loop would crash for some reason with half the batch enqueued. This meant we acquired some other "features", like being able to restart the process from a given ID (which would be helpfully logged when the process crashed).

### The new way ✨ 

- get `$batch_size` records from the table using `find_in_batches`
- per batch
  -  pluck all the IDs for the batch and give them as an array arg to a single `LoadEntityBatch` job

And in `LoadEntityBatch`:

- load the records for the IDs
- reduce them to a _single_ payload with `$batch_size` events in
- send that payload

### Advantages 📈 

- atomic: either the batch succeeds or it doesn't, so we eliminate dupes
- faster: fewer network connections
- much simpler outer loop so no need for all that recovery code

### Next 🔮 

The outer loop should now be fast enough to run in-process from a web request, so we could try to build a little UI on top of it!


.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 3.1.1
```
misaka, line 1: Any particular reason why not 3.1.2?

duncanjbrown: Maybe... because I wrote that commit at 11pm? 😆

spec/dfe/analytics/load_entities_spec.rb
```diff
@@ -1,11 +1,13 @@
...
 
 RSpec.describe DfE::Analytics::LoadEntities do
+  include ActiveJob::TestHelper
+
   before do
     allow(DfE::Analytics).to receive(:allowlist).and_return(
       {
         candidates: ['email_address']
-      }
+       }
```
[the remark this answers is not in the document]
misaka: image

misaka (approved):

PR #22, opened 2022-06-27
Don't crash on invalid UTF-8 in headers

duncanjbrown:
Users are able to craft request content which goes to BigQuery by setting request headers. These can include invalid `UTF-8`. This can cause issues at the point of job serialization (Sidekiq can't serialize the args) or at the point of sending to BQ (the Google Cloud library can't build the JSON).

We could preemptively roundtrip events to JSON and back again before putting them on the queue, but the Google library expects hashes so we can't reuse that JSON and it's wasteful. Instead, focus on the things we know can be malformed and fix those: ua and referer. We don't need to worry about about query params as Rails will reject those out of hand - see [ActionDispatch::Request::Utils.check_param_encoding](https://www.rubydoc.info/gems/actionpack/ActionDispatch%2FRequest%2FUtils.check_param_encoding)).

We decided to still send these events, because we send everything, but with � in place of unparseable bytes (which is how String#scrub replaces them).

### DIFF
--- a/lib/dfe/analytics/event.rb
+++ b/lib/dfe/analytics/event.rb
@@ -33,11 +33,11 @@ module DfE
       def with_request_details(rack_request)
         @event_hash.merge!(
           request_uuid: rack_request.uuid,
-          request_user_agent: rack_request.user_agent,
+          request_user_agent: ensure_utf8(rack_request.user_agent),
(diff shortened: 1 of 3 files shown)


misaka (approved):
🚢 💨

PR #23, opened 2022-06-28
convert data values using to_json

misaka:
When converting event data into the JSON payload to send to BQ, we used
to selectively only convert Hash values to JSON with "to_json". However,
some data values passed through for sending to BigQery may not present as a
Hash, e.g. trainees.progress in Register, causing inseration errors in
BQ.

This change makes us more lenient and will try to convert any value to
JSON if it implements the "to_json" method.

### DIFF
--- a/lib/dfe/analytics/event.rb
+++ b/lib/dfe/analytics/event.rb
@@ -101,6 +101,8 @@ module DfE
 
       def hash_to_kv_pairs(hash)
         hash.map do |(key, value)|
+          value = value.try(:as_json)
+
           if value.in? [true, false]
             value = value.to_s
           elsif value.is_a?(Hash)
--- a/spec/dfe/analytics/event_spec.rb
+++ b/spec/dfe/analytics/event_spec.rb
@@ -62,6 +62,17 @@ RSpec.describe DfE::Analytics::Event do
   end
 
   describe 'data pairs' do
+    let(:has_as_json_class) do
+      Struct.new(:colour, :is_cat) do
+        def as_json
+          {
+            colour: colour,
+            is_cat: is_cat
+          }
+        end
+      end
+    end
+
     it 'converts booleans to strings' do
       event = described_class.new
       output = event.with_data(key: true).as_json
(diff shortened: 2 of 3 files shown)


dfe-analytics.gemspec
```diff
@@ -36,6 +36,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rubocop-rspec', '~> 2'
   spec.add_development_dependency 'solargraph'
   spec.add_development_dependency 'sqlite3'
+  spec.add_development_dependency 'store_model'
```
duncanjbrown, line 39: Do we have to have this whole thing? Isn't the point the interface — ie that SomeObject implements to_json?

misaka:
One notable issue that this brings out is that if we send through an object with PII in it, it will be happily serialised and sent through to BQ. After we merge this PR, we should consider whether we need to add a layer of protection to ensure entities included as part of other entities only include the permitted fields.

dfe-analytics.gemspec
duncanjbrown: eg

```ruby
ObjectWithToJson = Struct.new(:colour) do
   def to_json
     JSON.generate({colour: colour})
   end
end
```

or something?

misaka: It's not quite `to_json`, I changed the implementation to check for `as_json`. We could hand-roll an `as_json` but I thought it would be better test to just use pre-built library. Happy to discuss.

duncanjbrown:
```ruby
ObjectWithAsJson = Struct.new(:colour) do
  def as_json
    {colour: colour}
  end
end
```

?

misaka:
One result of this change is that in the data sent by Register, dates will no longer have extra double-quotes around them, something originally introduced by an earlier attempt to get around the same problem this PR addresses. I don't think this will be an issue as the Dataform code optionally deals with extra double-quotes, but it's worth double-checking this.

dfe-analytics.gemspec
duncanjbrown: ^ imo this says "your object should support `as_json`", using the gem says "we support this particular gem", and I think I prefer the former?

misaka: Considering the spec is failing in CI because of this ... maybe hand-rolling will actually end up just being easier. _sigh_

misaka: I've adopted the Struct approach, thanks @duncanjbrown !

duncanjbrown (approved):

PR #26, opened 2022-06-28
Improve testing modes

thomasleese:
This makes two changes to the built-in testing modes to make them easier to use:

- Fixes a bug where `fake!` currently doesn't work as the stubbed `insert` method has the wrong signature.
- Adds a paragraph in the documentation explaining how to use the testing modes.

### DIFF
--- a/spec/dfe/analytics/send_events_spec.rb
+++ b/spec/dfe/analytics/send_events_spec.rb
@@ -45,6 +45,17 @@ RSpec.describe DfE::Analytics::SendEvents do
       end
     end
 
+    context 'when using fake testing mode' do
+      it 'does not go out to the network' do
+        request = stub_analytics_event_submission
+
+        DfE::Analytics::Testing.fake! do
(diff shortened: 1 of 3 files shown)


spec/dfe/analytics/send_events_spec.rb
```diff
@@ -45,6 +45,17 @@
       end
     end
 
+    context 'when using fake testing mode' do
+      it 'does not go out to the network' do
+        request = stub_analytics_event_submission
+
+        DfE::Analytics::Testing.fake! do
```
duncanjbrown, line 52: This block isn't necessary as it's the default state, set in the body of the `Testing` module (also a la sidekiq) — could rename spec and/or add something to the docs?

thomasleese: Yeah it's true that this is default, but I wanted this test to be explicit in case that changes in the future. It's clear that whatever might change around this test, we're at least using the `fake` testing mode in it.

PR #29, opened 2022-07-13
Raise insertion errors from BigQuery

thomasleese:
When inserting data in to BigQuery we need to inspect the response for errors as although it may have a 200 HTTP status code, schema-related errors that occur on insertion are included in the error information of the response.

See https://cloud.google.com/bigquery/docs/error-messages#streaming-success for more information on this.

[Trello Card](https://trello.com/c/WFVVrgpr/740-raise-insertion-errors-in-bigquery)


duncanjbrown:
🎉 Works very well and causes retries! The log output is still confusingly positive with "Success" and "200" messages — perhaps we can improve that in future.

Screenshot 2022-07-14 at 13 58 38

Re the `@insert_errors` variable, we're not trapping these anywhere in the gem, so it goes unused — I don't think Sentry can pick it up? Maybe a future improvement to aid quick inspection in error reports could be to munge the `message` fields from the Google error objects into a string and use that as the exception message?

duncanjbrown (approved):
great stuff, thanks @thomasleese  — approving with one comment about `@insert_errors`

PR #30, opened 2022-07-19
Refactor to straighten out language and improve dfe:analytics:check

duncanjbrown:
The useful `analytics:check` task detects when we add and remove fields from applications, but it had a shortcoming: it couldn't tell when a field existed in both the allowlist and the blocklist. This wasn't considered first time around because _why would that ever happen_. We now know that it _does_ happen, and it happens when an entity wasn't previously in the allowlist but was in the blocklist (because we ignored it), and now we've added it to the allowlist (because we've decided not to ignore it) and now it's appearing in both.

Not necessarily a problem, but it's unprofessional and confusing and it looks bad so it's no longer allowed 🚫 .

This PR also contains some refactoring to the README and methods which refer to the word `model`. We don't believe in `models` out here; we believe in `entities`. Keep usage of `model` and `table_name` to an absolute minimum so people don't get confused when we're referring to three somewhat equivalent things by three different names.

🤔 **You may notice** that whilst this PR makes a point of saying that we send entity events rather than model events, the README still only talks about mixing the `Entities` behaviour into models. This is because the _next_ PR is going to do away with the manual mixin altogether following very good idea from @misaka that we could do it automatically with information we've got lying around in `DfE::Analytics` — progress continues on that [here](https://github.com/DFE-Digital/dfe-analytics/compare/model-to-entity?expand=1). This will fix a bug where we don't send updates for join tables without an associated model.

This PR best reviewed commit by commit.

### DIFF
--- a/lib/dfe/analytics/fields.rb
+++ b/lib/dfe/analytics/fields.rb
@@ -23,6 +23,16 @@ module DfE
         diff_model_attributes_against(allowlist)[:surplus]
       end
 
+      def self.conflicting_fields
+        allowlist.keys.reduce({}) do |conflicts, entity|
+          intersection = Array.wrap(blocklist[entity]) & allowlist[entity]
+
+          conflicts[entity] = intersection if intersection.any?
+
+          conflicts
+        end
+      end
+
       def self.diff_model_attributes_against(*lists)
         DfE::Analytics.all_entities_in_application
           .reduce({ missing: {}, surplus: {} }) do |diff, entity|
--- a/lib/dfe/analytics/fields.rb
+++ b/lib/dfe/analytics/fields.rb
@@ -11,47 +11,52 @@ module DfE
         DfE::Analytics.allowlist
       end
 
+      def self.database
+        DfE::Analytics.all_entities_in_application
+          .reduce({}) do |list, entity|
+            attrs = DfE::Analytics.model_for_entity(entity).attribute_names
+            list[entity] = attrs
+
+            list
+          end
+      end
+
       def self.generate_blocklist
-        diff_model_attributes_against(allowlist)[:missing]
+        diff_between(database, allowlist)
       end
 
       def self.unlisted_fields
-        diff_model_attributes_against(allowlist, blocklist)[:missing]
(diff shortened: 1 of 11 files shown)


lib/dfe/analytics/fields.rb
```diff
@@ -24,25 +24,21 @@ def self.surplus_fields
...
+            end.reduce(:concat)
+
+            model = DfE::Analytics.model_for_entity(entity)
+
+            missing_attributes = model.attribute_names - attributes_considered
+            surplus_attributes = attributes_considered - model.attribute_names
+
+            diff[:missing][entity] = missing_attributes if missing_attributes.any?
+
+            diff[:surplus][entity] = surplus_attributes if surplus_attributes.any?
```
misaka, line 51: I think if you're going to be refactoring this bit, you may as well reduce the complexity here too. Maybe something like (note, these are untested sketches): 

```ruby
      def entity_attributes_from_lists(entity, lists)
        lists.map do |list|
          # for each list of model attrs, look up the attrs for this model
          list.fetch(entity, [])
        end.reduce(:concat)
      end

      def self.entities_with_missing_attributes_from_lists(*lists)
        DfE::Analytics.all_entities_in_application
          .reduce({}) do |entities_missing_attributes, entity|
            attributes_considered = entity_attributes_from_lists(entity, lists)
            model = DfE::Analytics.model_for_entity(entity)
            entities_missing_attributes[entity] = model.attribute_names - attributes_considered
          end.reject { |_entity, attributes| attributes.empty? }
      end

      def self.entities_with_surplus_attributes_from_lists(*lists)
        DfE::Analytics.all_entities_in_application
          .reduce({}) do |entities_surplus_attributes, entity|
            attributes_considered = entity_attributes_from_lists(entity, lists)
            model = DfE::Analytics.model_for_entity(entity)
            entities_surplus_attributes[entity] = model.attribute_names - attributes_considered
          end.reject { |_entity, attributes| attributes.empty? }
      end
```

misaka (changes requested):
Looks good! 🚢

misaka (approved):
Argh. Wrong circle. 🤦

PR #32, opened 2022-07-28
v1.3.0

misaka:
Looks like the runtime dependencies from the `gemspec` were updated as part of this. This looks surprising, but it only effects our local dev env, each project will pull in their own versions of these files. The specs are still all 🟢 with this update.


tvararu (approved):
:ship:

thomasleese (approved):

PR #34, opened 2022-08-01
[770] handle arrays of JSON-friendly objects

misaka:
We've previously added special handling to event data processing for objects that can be converted to JSON; this change expands that to handle an array of these objects that can be converted to JSON. Words are hard, a ~~picture~~JSON structure is worth a thousand words:

BEFORE BAD:

```json
        {
          "key": "accrediting_provider_enrichments",
          "value": [
            {
              "UcasProviderCode": "N21",
              "Description": "",
              "validation_context": null,
              "errors": {
              }
            }
          ]
        }
```

NOW GOOD:

```json
        {
          "key": "accrediting_provider_enrichments",
          "value": [
            "{\"UcasProviderCode\":\"N21\",\"Description\":\"\",\"validation_context\":null,\"errors\":{}}"
          ]
        }
```

(we now encode each element of the array to JSON).

This has change has also been tested in the test environment.

### DIFF
--- a/lib/dfe/analytics/event.rb
+++ b/lib/dfe/analytics/event.rb
@@ -99,17 +99,25 @@ module DfE
 
       private
 
-      def hash_to_kv_pairs(hash)
-        hash.map do |(key, value)|
-          value = value.try(:as_json)
+      def convert_value_to_json(value)
+        value = value.try(:as_json)
 
-          if value.in? [true, false]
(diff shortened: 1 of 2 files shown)


misaka:
cc: @avinhurry for the assist

duncanjbrown (approved):

PR #36, opened 2022-08-02
Adopt with_model

duncanjbrown:
I found this extremely useful gem which lets us declare the app's models at the top of specs instead of maintaining them as state in the dummy app 🙌 .

Migrate existing specs to it. I'll rebase the auto-init branch on top of this.


misaka (commented):
I really like this `with_model` thing, _however_, a couple concerns.

1. I was wondering if the right way for us to check end-to-end integration with BigQuery is to use the dummy app that gets superceded by `with_model` here.
2. We've removed the dummy apps models, but the controllers stay there. It could be nice to keep the dummy app as a more complete example of how to setup an app, which could integrate nicely the integration tests mentioned above.

So ... options:

1. Use the dummy app for these purposes.
2. Setup a separate dummy app for integration testing that also demonstrates how things are setup.

Also, is there a way to make models without requiring ActiveRecord? Something like [active_attr](https://github.com/cgriego/active_attr)? Although I think our other PR more closely ties us to ActiveRecord so maybe not.

duncanjbrown:
As discussed:

- We're going to clean up the models in dummy app so they're not dangling
- We're going to try and add a proper e2e integration test that goes all the way to BQ, for running in CI with suitable creds

duncanjbrown:
updated!

misaka (approved):

PR #37, opened 2022-08-03
Initialize model callbacks on boot

duncanjbrown:
See commits.

Manually mixing the behaviour in to ActiveRecord classes does not achieve what we want to achieve: it missed out HABTM relations. So hook up everything automatically — `analytics.yml` will determine what gets sent. As part of this piece of work we now refuse to boot if that file isn't correct, which should surface configuration issues as early as possible.

Depends on #36.

### DIFF
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -17,6 +17,8 @@ require 'dfe/analytics/railtie'
 
 module DfE
   module Analytics
+    class ConfigurationError < StandardError; end
+
     def self.events_client
       @events_client ||= begin
         require 'google/cloud/bigquery'
@@ -28,7 +30,7 @@ module DfE
           bigquery_api_json_key
         ].select { |val| config.send(val).nil? }
 
-        raise "DfE::Analytics: missing required config values: #{missing_config.join(', ')}" if missing_config.any?
+        raise(ConfigurationError, "DfE::Analytics: missing required config values: #{missing_config.join(', ')}") if missing_config.any?
 
         Google::Cloud::Bigquery.new(
           project: config.bigquery_project_id,
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -76,6 +76,10 @@ module DfE
       config.queue                 ||= :default
     end
 
+    def self.initialize!
+      DfE::Analytics::Fields.check!
+    end
+
     def self.enabled?
       config.enable_analytics.call
     end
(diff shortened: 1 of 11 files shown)


lib/dfe/analytics.rb
```diff
@@ -74,6 +76,22 @@ def self.configure
...
+    def self.initialize!
+      DfE::Analytics::Fields.check!
+
+      @shown_deprecation_warning = false
+
+      entities_for_analytics.each do |entity|
+        model = model_for_entity(entity)
+        if model.include?(DfE::Analytics::Entities) && !@shown_deprecation_warning
+          Rails.logger.info("DEPRECATION WARNING: DfE::Analytics::Entities was manually included in a model (#{model.name}), but it's included automatically since v1.4. You're running v#{DfE::Analytics::VERSION}. To silence this warning, remove the include from model definitions in app/models.")
+          @shown_deprecation_warning = true
```
misaka, line 88: It's tempting to include a deprecation warning for every model that does this. But, then, if it's only included in `ApplicationRecord` then that might be annoying.

I'd go for annoy and apologise later. 😛

lib/dfe/analytics.rb
```diff
@@ -140,6 +158,16 @@ def self.entity_model_mapping
...
       @entity_model_mapping ||= begin
+        # Gems like devise put helper methods into controllers, and they add
+        # those methods at the time the routes file is parsed.
+        #
+        # Rails.configuration.eager_load = true will cause routes to be loaded
+        # before controllers; a direct call to Rails.application.eager_load! will
+        # not. To avoid this specific conflict with devise and possibly other
+        # gems/engines, proactively load the routes unless
+        # configuration.eager_load is set.
+        Rails.application.reload_routes! unless Rails.configuration.eager_load
```
misaka, line 169: Did we only need to do this in `development` env? Or do we not need to check that because in non-dev the routes will be loaded at this point anyway? I don't quite remember what we said, but might be worth explaining why this is ok to do in production if there's anything  interesting to say.

misaka:
Might be worth baseing this PR onto the `with_model` PR, just to keep it's diff set clean?

lib/dfe/analytics.rb
duncanjbrown: I agree, let's be nuisances

lib/dfe/analytics.rb
duncanjbrown: Will beef up this comment to explain what default configs there are in prod/dev

duncanjbrown:
OK: all updated

misaka (approved):

PR #38, opened 2022-08-10
Fix rake tasks

darokel:
Tried to run one of the rake tasks in Publish and it was complaining about structure of args. This PR introduces a couple of fixes:

- Fixes rake task `:import_all_entities`, eg:

```ruby
NameError: undefined local variable or method `entities_for_analytics' for main:Object

      entities_for_analytics.each do |entity_name|

```

- Fixes rake task `:import_entity` to correctly destructure args otherwise we get:

```ruby
ArgumentError: wrong number of arguments (given 1, expected 0; required keyword: entity_name)
```

- Pins async gem to a version that hasn't been yanked

Also pinned the async gem to a version that hasn't been yanked, let me know if this is ok.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -76,7 +76,7 @@ GEM
     addressable (2.8.0)
       public_suffix (>= 2.0.2, < 5.0)
     ast (2.4.2)
-    async (2.0.3)
+    async (2.0.0)
       console (~> 1.10)
       io-event (~> 1.0.0)
       timers (~> 4.1)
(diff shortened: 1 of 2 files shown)


misaka:
Just to add context, `apply` and `register` are still on Ruby 2.7.x, which is why despite the Ruby version being set to `3.1.2` in `.tool-versions`, we never encountered this error. The are also no tests for that slim bit of the code in the task.

misaka (approved):
🚀

PR #42, opened 2022-08-18
allow user id field to be configurable

misaka:
Some systems may not be able to, or prefer not to, use the id attribute
on user to identify the user. Allow these systems to call a different
method on the user to return an identifier.

This change should not break existing systems as defaults to using `user.id` as the previous systems did.


ethax-ross (approved):
Looks good to me 💯

thomasleese (approved):

duncanjbrown (approved):

PR #45, opened 2022-09-01
Make backfill resilient

duncanjbrown:
Objectives:

- don't overflow queues
- don't hit BQ byte limits

Hopefully this isn't too complicated.

Still to do: find out the actual byte limit and set it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 /spec/dummy/storage/
 /spec/dummy/tmp/
 .rspec_status
+.yardoc
--- a/lib/dfe/analytics/event.rb
+++ b/lib/dfe/analytics/event.rb
@@ -97,6 +97,10 @@ module DfE
         self
       end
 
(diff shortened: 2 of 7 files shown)


misaka:
Where is this 10k limit coming from?

It's just that this approach seems reasonable if we have a vague limit of about 10k, but if this _is_ a vague limit then what is the limitation, really?

duncanjbrown:
> Where is this 10k limit coming from?

I made it up! See description. I want to experiment to confirm the actual limit — docs haven't been helpful.

misaka:
What errors are we seeing?

duncanjbrown:
OK I've simplified this on the basis of https://cloud.google.com/bigquery/quotas#streaming_inserts

new routine:

break everything up into 500 row bits as suggested by google
measure size before sending, and if we end up going over 10mb, split the batch

misaka (approved):

PR #47, opened 2022-09-16
Change ActiveJob log level

duncanjbrown:
It was necessary to defer setting the logger to after ActiveJob had loaded, because ActiveJob sets a logger in its own Railtie.

https://github.com/rails/rails/blob/a5018cf149d47e1aed1e74dd35e4d519b507c364/activejob/lib/active_job/railtie.rb#L13-L15

### DIFF
--- a/lib/dfe/analytics/railtie.rb
+++ b/lib/dfe/analytics/railtie.rb
@@ -13,6 +13,19 @@ module DfE
         app.config.middleware.use DfE::Analytics::Middleware::RequestIdentity
       end
 
+      initializer 'dfe.analytics.logger' do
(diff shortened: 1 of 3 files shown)


saliceti (approved):

misaka:
Is there any value to setting the logging level to `error` or `critical`?

duncanjbrown:
Interesting idea

misaka:
> Interesting idea

Just thinking it would still be good to see `error` or `fatal` messages. Even `warning` would be enough, as the noisy logging is being logged as `info`.

saliceti (approved):

slorek (approved):

PR #51, opened 2022-09-26
Quietly fail init if no database is present

duncanjbrown:
Failing loudly under these circs context isn't helpful because it doesn't reflect a problem with the analytics setup; apps in CI often run the asset precompilation step without a database present.


duncanjbrown:
This has built successfully on Publish (https://github.com/DFE-Digital/publish-teacher-training/actions/runs/3134716188/jobs/5089961606) so opening for review

misaka (approved):
Looks good.

PR #53, opened 2022-09-28
Ignore models with non-id primary keys

duncanjbrown:
We can probably handle these, but it would involve fixing find_in_batches to work with non-id PKs, which is out of scope for now


misaka (approved):
Agreed. 👞🥫

PR #54, opened 2022-09-28
🍀 Four kinds of init error

duncanjbrown:
4️⃣   error conditions we discovered during recent rollout in Publish and Apply CI pipelines.

- ActiveRecord::SchemaMigrations table is being considered for dfe-analytics (ignore it)
- discovering migrations pending on dfe-analytics init
- discovering missing database connection when database not present, also on init
- loading models (in init) before connecting to DB

The final example is important because some contexts which lack db connections (eg asset precompilation) also don't provide all necessary eg ENV vars which would be required were we loading the whole app (which this init process does).

So discover early if we have no database (call `ActiveRecord::Base.connection`). & if anything happens during Fields.check! catch that too. Log lines will tell us if we're bailing for want of a DB or for want of a migration. (There's no point in running check! if migration yet to run).

### DIFF
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -79,13 +79,7 @@ module DfE
     end
 
     def self.initialize!
-      begin
-        ActiveRecord::Base.connection
-      rescue ActiveRecord::ActiveRecordError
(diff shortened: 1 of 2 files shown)


misaka (approved):
💯

PR #60, opened 2023-01-12
Allowing pushing custom events

BroiSatse:
This small change allows all the users to define the custom events they would like to push to analytics database.


lib/dfe/analytics.rb
```diff
@@ -119,6 +119,12 @@ def self.blocklist
       Rails.application.config_for(:analytics_blocklist)
     end
 
+    def self.custom_events
+      Rails.application.config_for(:analytics_custom_events)
+    rescue RuntimeError
```
duncanjbrown, line 124: Bad Rails not returning specific exception type in this context!

README.md
```diff
@@ -177,6 +178,33 @@ DfE::Analytics.config.queue = :default
 
 Alternatively you may consider setting the priority of the jobs according to your chosen ActiveJob adapter's conventions.
 
+### 8. Custom events
+
+If you wish to send custom analytics event, create a file `config/analytics_custom_events.yml` containing an array of your custom events types under a `shared` key like:
```
duncanjbrown, line 183:
```suggestion
If you wish to send custom analytics events, create a file `config/analytics_custom_events.yml` containing an array of your custom event types under a `shared` key like:
```

duncanjbrown (commented):
Looks great, thank you! (And docs too ❤️). Please could you add a test proving that you can pass an arbitrary custom event as long as it appears in the file?

lib/dfe/analytics.rb
duncanjbrown: It's painful, but I think I'd rather live with generic exception than break the abstraction of `config_for`? wdyt?

duncanjbrown (approved):
Fab, thanks @BroiSatse — pls merge once Rubocop happy!

ddippolito:
@duncanjbrown can you merge this for us please?

PR #61, opened 2023-01-23
Custom events

BroiSatse:
DfE::Analytics::SendEvents expects an array of hashes. This change allows passing constructed DfE::Analytics::Event object as well, without need to convert them into hashes first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ event = DfE::Analytics::Event.new
 Once all the events have been constructed, simply send them to your analytics:
 
 ```ruby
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -202,7 +202,7 @@ event = DfE::Analytics::Event.new
 Once all the events have been constructed, simply send them to your analytics:
 
 ```ruby
-DfE::Analytics::SendEvent.do(event, event2, event3, ...)
+DfE::Analytics::SendEvents.do([event, event2, event3])
```
duncanjbrown, line 205: 🙏

duncanjbrown (approved):

PR #63, opened 2023-02-07
Log Matched Events

asatwal:
[Trello-975](https://trello.com/c/rUufn40w/975-investigate-silent-failures-in-dfe-analytics)

Create an events filter for logging, so that we log only the events of interest. This will allow targeted logging for diagnostic and debugging purposes.

The events filter is held in a YAML config.

Event filter example:

``` Yaml
#
# Allows logging of events that match the type and given filters on the fields
#
# All values are converted to regular expressions for matching
#
# Any filter fields can be defined as long as the field exists in the target event
#
# If there are mutiple filters then at least one must match the event
#
# All filter fields must match the event fields for a filter to match
#
shared:
  event_filters:
    -
      type: (create|update|delete)_entity
      entity_table_name: course_options
      data:
        key: id
        value: 12345
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,9 +348,9 @@ Event filters allow targeted event logging for diagnostic and debug purposes. Th
 When defining event filters, note the following:
 - All values are converted to regular expressions for matching
 - Any filter fields can be defined as long as the field exists in the target event
-- A filter must be a hash and nesteds field are allowed.
-- If a corresponding hash field in the target event is not found, then the remaining value in the target is converted into a string and compared with the value from the filter. The remaining nested fileds in the filter are then ignored. This may result in a wider match than expected
-- If there are mutiple filters then at least one must match the event
+- A filter must be a hash and nested fields are allowed
+- If a corresponding hash field in the target event is not found, then the remaining value in the target is converted into a string and compared with the value from the filter. The remaining nested fields in the filter are then ignored. This may result in a wider match than expected
+- If there are multiple filters then at least one must match the event
 - All filter fields must match the event fields for a filter to match
 
 In the above example, all CRUD entity events to the `course_options` table and `id` matching value `1234` will be logged, or any update events to the `courses` table will also be logged.
--- a/README.md
+++ b/README.md
@@ -334,12 +334,13 @@ If you wish to log events for debug purposes, create a file `config/analytics_ev
 shared:
   event_filters:
     -
-      type: (create|update|delete)_entity
+      event_type: (create|update|delete)_entity
       entity_table_name: course_options
-      key: id
-      value: 12345
+      data:
+        key: id
+        value: 12345
     -
-      type: update_entity
+      event_type: import_entity
       entity_table_name: courses
 ```
 
@@ -353,7 +354,7 @@ When defining event filters, note the following:
 - If there are multiple filters then at least one must match the event
 - All filter fields must match the event fields for a filter to match
 
-In the above example, all CRUD entity events to the `course_options` table and `id` matching value `1234` will be logged, or any update events to the `courses` table will also be logged.
+In the above example, all create, delete or update entity events to the `course_options` table and `id` matching value `1234` will be logged, or any import entity events to the `courses` table will also be logged.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -325,6 +325,36 @@ To reimport just one entity, run:
 bundle exec rails dfe:analytics:import_entity[entity_name]
 ```
 
+
+## Event debugging
+
+If you wish to log events for debug purposes, create a file `config/analytics_event_debug.yml` containing an array of your event filters  under a `shared` key like:
+
+```yaml
+shared:
+  event_filters:
+    -
+      type: (create|update|delete)_entity
+      entity_table_name: course_options
+      key: id
+      value: 12345
+    -
+      type: update_entity
+      entity_table_name: courses
+```
+
+Event filters allow targeted event logging for diagnostic and debug purposes. The logging level is `info`.
+
+When defining event filters, note the following:
+- All values are converted to regular expressions for matching
+- Any filter fields can be defined as long as the field exists in the target event
+- A filter must be a hash and nesteds field are allowed.
+- If a corresponding hash field in the target event is not found, then the remaining value in the target is converted into a string and compared with the value from the filter. The remaining nested fileds in the filter are then ignored. This may result in a wider match than expected
+- If there are mutiple filters then at least one must match the event
+- All filter fields must match the event fields for a filter to match
+
+In the above example, all CRUD entity events to the `course_options` table and `id` matching value `1234` will be logged, or any update events to the `courses` table will also be logged.
+
 ## Contributing
 
 1. Make a copy of this repository
--- a/README.md
+++ b/README.md
@@ -350,11 +350,39 @@ When defining event filters, note the following:
 - All values are converted to regular expressions for matching
 - Any filter fields can be defined as long as the field exists in the target event
 - A filter must be a hash and nested fields are allowed
-- If a corresponding hash field in the target event is not found, then the remaining value in the target is converted into a string and compared with the value from the filter. The remaining nested fields in the filter are then ignored. This may result in a wider match than expected
+- If a corresponding hash field in the target event is not found, then the remaining value in the target is converted into a string and compared with the value from the filter. The remaining nested fields in the filter are then ignored. This may result in a wider match than expected. Please see section on matching for non hash fields below
 - If there are multiple filters then at least one must match the event
 - All filter fields must match the event fields for a filter to match
 
 In the above example, all create, delete or update entity events to the `course_options` table and `id` matching value `1234` will be logged, or any import entity events to the `courses` table will also be logged.
+
+### Matching on non-hash fields
+
+This is best demonstrated by example.
+
+Given the above event filters and the following target event:
+
+``` Ruby
+  {
+    'entity_table_name' => 'course_options',
+    'event_type' => 'update_entity',
+    'data' => [
+      { 'key' => 'id', 'value' => ['12345'] },
+      { 'key' => 'course_id', 'value' => ['42'] }
+    ]
+  }
+```
+
+Then on matching, there is a one to one correspondence on the `entity_table_name` and `event_type` fields, so these match OK. However, in the target event `data` field there is no hash value, so the `key` field with value of `id` is compared with the whole of the target `data` field converted to a string, and the `value` field with value of `12345` would also be compared with the whole of the target `data` field.
+
+So the comparisons in Ruby would be:
+
+``` Ruby
+  /id/ =~ "[{ 'key' => 'id', 'value' => ['12345'] }, { 'key' => 'course_id', 'value' => ['42'] }]"
+  /12345/ =~ "[{ 'key' => 'id', 'value' => ['12345'] }, { 'key' => 'course_id', 'value' => ['42'] }]"
+```
+
+The fields do match successfully, but note the the first comparison matches `id` on `id` and `course_id` so the match would be wider than expected in some instances.
 
 ## Contributing
 
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -6,6 +6,7 @@ require 'dfe/analytics/event_schema'
 require 'dfe/analytics/fields'
 require 'dfe/analytics/entities'
 require 'dfe/analytics/event'
+require 'dfe/analytics/event_matcher'
 require 'dfe/analytics/analytics_job'
 require 'dfe/analytics/send_events'
 require 'dfe/analytics/load_entities'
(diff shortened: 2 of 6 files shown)


spec/dfe/analytics/event_matcher_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+
+    describe '.matched?' do
+      context 'when the table name and event type matches' do
+        let(:logging) do
+          {
+            event_filters: [
+              {
+                event_type: 'update_entity',
+                entity_table_name: 'application_choice_details'
```
stevenleggdfe, line 23: @asatwal I appreciate it's just a test but application_choice_details isn't an entity_table_name in any of our databases - worth changing for clarity?

asatwal: Yes, it's worth changing. I was trying to think of a realistic example, but my application knowledge isn't that great. What would be the entity_table_name for the current problems we are experiencing with the nightly job ?

stevenleggdfe: course_options I think

asatwal: Updated to a more realistic example.

README.md
```diff
@@ -325,6 +325,36 @@ To reimport just one entity, run:
...
+```
+
+Event filters allow targeted event logging for diagnostic and debug purposes. The logging level is `info`.
+
+When defining event filters, note the following:
+- All values are converted to regular expressions for matching
+- Any filter fields can be defined as long as the field exists in the target event
+- A filter must be a hash and nesteds field are allowed.
+- If a corresponding hash field in the target event is not found, then the remaining value in the target is converted into a string and compared with the value from the filter. The remaining nested fileds in the filter are then ignored. This may result in a wider match than expected
+- If there are mutiple filters then at least one must match the event
```
stevenleggdfe, line 353: @asatwal typo: mutiple

README.md
```diff
@@ -325,6 +325,36 @@ To reimport just one entity, run:
...
+      type: update_entity
+      entity_table_name: courses
+```
+
+Event filters allow targeted event logging for diagnostic and debug purposes. The logging level is `info`.
+
+When defining event filters, note the following:
+- All values are converted to regular expressions for matching
+- Any filter fields can be defined as long as the field exists in the target event
+- A filter must be a hash and nesteds field are allowed.
```
stevenleggdfe, line 351: @asatwal typo: nesteds field

README.md
```diff
@@ -325,6 +325,36 @@ To reimport just one entity, run:
...
+
+## Event debugging
+
+If you wish to log events for debug purposes, create a file `config/analytics_event_debug.yml` containing an array of your event filters  under a `shared` key like:
+
+```yaml
+shared:
+  event_filters:
+    -
+      type: (create|update|delete)_entity
```
stevenleggdfe, line 337: @asatwal what about import_entity events? Would they be supported too?

README.md
```diff
@@ -325,6 +325,36 @@ To reimport just one entity, run:
 bundle exec rails dfe:analytics:import_entity[entity_name]
 ```
 
+
+## Event debugging
+
+If you wish to log events for debug purposes, create a file `config/analytics_event_debug.yml` containing an array of your event filters  under a `shared` key like:
+
+```yaml
+shared:
```
stevenleggdfe, line 334: @asatwal do we need the shared key or could event_filters be the top level?

README.md
```diff
@@ -325,6 +325,36 @@ To reimport just one entity, run:
...
+      entity_table_name: courses
+```
+
+Event filters allow targeted event logging for diagnostic and debug purposes. The logging level is `info`.
+
+When defining event filters, note the following:
+- All values are converted to regular expressions for matching
+- Any filter fields can be defined as long as the field exists in the target event
+- A filter must be a hash and nesteds field are allowed.
+- If a corresponding hash field in the target event is not found, then the remaining value in the target is converted into a string and compared with the value from the filter. The remaining nested fileds in the filter are then ignored. This may result in a wider match than expected
```
stevenleggdfe, line 352: @asatwal typo: fileds

stevenleggdfe (changes requested):

README.md
asatwal: `shared` in this context means shared between environments. So this filter definition will be applicable to development, test and production.

README.md
asatwal: So type can be absolutely anything. It's actually converted to a regular expressions (Which is the first bullet point: `All values are converted to regular expressions for matching`). So if we wanted to match on `import_entity`, we would just add another filter to the event_filters array in the YAM file:

``` YAML
  event_filters:
    -
      type: import_entity
```

README.md
asatwal: thanks @stevenleggdfe - I've actually now turned the spell checker on in my mardown editor :-)

README.md
asatwal: Fixed

README.md
asatwal: Fixed

README.md
```diff
@@ -325,6 +325,36 @@ To reimport just one entity, run:
...
+If you wish to log events for debug purposes, create a file `config/analytics_event_debug.yml` containing an array of your event filters  under a `shared` key like:
+
+```yaml
+shared:
+  event_filters:
+    -
+      type: (create|update|delete)_entity
+      entity_table_name: course_options
+      key: id
+      value: 12345
```
duncanjbrown, line 340: I find this example confusing because these key/value pairs actually refer to the contents of the nested `data` hash so the structure of this is not similar to the structure of an actual event. Would it be better to explicitly nest them under a `data` key? (I think this is what the code actually does in the spec!). Also, what would the process be for adding multiple key/value pairs? ([Event schema, for reference](https://github.com/DFE-Digital/dfe-analytics/blob/main/config/event-schema.json))

I wonder if this PR could be simplified by simply restricting to top-level string fields like `type`, `entity_table_name` etc, and then we could loop back and layer on `request_query` and `data` options?

README.md
```diff
@@ -325,6 +325,36 @@ To reimport just one entity, run:
...
+      entity_table_name: courses
+```
+
+Event filters allow targeted event logging for diagnostic and debug purposes. The logging level is `info`.
+
+When defining event filters, note the following:
+- All values are converted to regular expressions for matching
+- Any filter fields can be defined as long as the field exists in the target event
+- A filter must be a hash and nested fields are allowed
+- If a corresponding hash field in the target event is not found, then the remaining value in the target is converted into a string and compared with the value from the filter. The remaining nested fields in the filter are then ignored. This may result in a wider match than expected
```
duncanjbrown, line 352: I'm struggling to understand this condition. Could we simplify somehow? (perhaps by removing this behaviour?)

README.md
```diff
@@ -325,6 +325,36 @@ To reimport just one entity, run:
...
+- If there are multiple filters then at least one must match the event
+- All filter fields must match the event fields for a filter to match
+
+In the above example, all CRUD entity events to the `course_options` table and `id` matching value `1234` will be logged, or any update events to the `courses` table will also be logged.
```
duncanjbrown, line 356: We should probably clarify this to take into account @stevenleggdfe's point above — there's no `R` in the `CRUD`, but `import_entity` would add it.

lib/dfe/analytics.rb
```diff
@@ -135,6 +142,10 @@ def self.log_only?
       config.log_only
     end
 
+    def self.event_debug_enabled?
```
duncanjbrown, line 145: 👌

lib/dfe/analytics.rb
```diff
@@ -127,6 +128,12 @@ def self.custom_events
       []
     end
 
+    def self.event_debug
```
duncanjbrown, line 131: Maybe `event_debug_filters`, so it's clear that this isn't a switch?

lib/dfe/analytics/event_matcher.rb
```diff
@@ -0,0 +1,52 @@
...
+        regexp = Regexp.new(filter_value)
+
+        regexp.match?(event_value)
+      end
+
+      def event_value_for(nested_fields)
+        # If nested hash fields in a filter don't correspond to hashes in the event THEN
+        # - Convert the remaining value into a string (note: this maybe a whole array)
+        # - Don't dig any deeper into the event on the first non hash value
+        # - Will result in greedy and overzealous match as whole of nested structure compared
```
duncanjbrown, line 43: There's no test for this subtle behaviour (and as far as I can see, no immediate need)

spec/dfe/analytics/send_events_spec.rb
```diff
@@ -65,6 +83,58 @@
...
+              {
+                request_method: 'POST',
+                request_path: '/provider/applications',
+                namespace: 'provider_interface'
+              }
+            ]
+          }
+        end
+
+        it 'does not logs the event' do
```
duncanjbrown, line 131:
```suggestion
        it 'does not log the event' do
```

duncanjbrown (commented):
Great! Comments inline. I think there are a couple more things to think about re nested fields, and I'm not convinced we need to support them at this stage (unless something has changed), so I've proposed keeping this PR minimal and focussed on the use case at hand, ie matching top-level strings in the event payload.

README.md
stevenleggdfe: @asatwal that make sense. But in that case perhaps this behaviour should be documented if it isn't already?

duncanjbrown: It's core Rails behaviour! Agree it's a bit funky.

README.md
asatwal: @duncanjbrown  - You are absolutely correct, key and value should be nested under data. The type key is also wrong! Sorry - my bad - I've updated now.

Multiple values for key/value pairs would require another filter. For example if we wanted to search for field `course_id` value `42` then
in the above example then we would add another filter to the array:

``` YAML
    -
      event_type: (create|update|delete)_entity
      entity_table_name: course_options
      data:
        key: course_id
        value: 42
```

The reason why I've done nesting like this is to keep the code generic. The code does not refer to any fields by name and does a match on top level fields and then nested fields by recursing over them.

README.md
asatwal: So this is probably best demonstrated by example. So given the example filter:

``` yaml
shared:
  event_filters:
    -
      event_type: (create|update|delete)_entity
      entity_table_name: course_options
      data:
        key: id
        value: 12345
```

Also the target event:

``` Ruby
  {
    'entity_table_name' => 'course_options',
    'event_type' => 'update_entity',
    'data' => [
      { 'key' => 'id', 'value' => ['12345'] },
      { 'key' => 'course_id', 'value' => ['42'] }
    ]
  }
```

Then on matching, there is a one to one correspondence on the `entity_table_name` and `event_type` fields, so thees matches are OK.
However, in the `data` field there is no hash value, so the `key` field with value of `id` is compared with the whole of the target data field converted to a string, and the `value` field with value of `['12345']` would also be compared.

So the comparisons would be:

``` Ruby
  /id/ =~ "[{ 'key' => 'id', 'value' => ['12345'] }, { 'key' => 'course_id', 'value' => ['42'] }]"
  /\['12345'\]/ =~ "[{ 'key' => 'id', 'value' => ['12345'] }, { 'key' => 'course_id', 'value' => ['42'] }]"

```

Matching nested fields in a hash generically is going to be non-trivial, and I'm struggling to explain this in a simple fashion, but as you point out this is not necessarily required for the problem in hand and maybe best to keep it simple.

README.md
asatwal: Indeed. I shouldn't have used `CRUD` here - Replaced with `create, delete or update`. Also replaced the second example with @stevenleggdfe's question on `import_entity` event type.

lib/dfe/analytics.rb
asatwal: Fixed.

lib/dfe/analytics/event_matcher.rb
asatwal: So given the example above about the key/value matching in the `data` field this might make more sense.

Tests for nested hash key is here (matching):
https://github.com/DFE-Digital/dfe-analytics/pull/63/files#diff-ea37205c2d841767ef77850589cb77f297a0cc862dd3b01acbc07e188bea98f6R33

Tests for nested hash key is here (not matching):
https://github.com/DFE-Digital/dfe-analytics/pull/63/files#diff-ea37205c2d841767ef77850589cb77f297a0cc862dd3b01acbc07e188bea98f6R53

asatwal:
> Great! Comments inline. I think there are a couple more things to think about re nested fields, and I'm not convinced we need to support them at this stage (unless something has changed), so I've proposed keeping this PR minimal and focussed on the use case at hand, ie matching top-level strings in the event payload.

@duncanjbrown - Thanks for the review. Some very useful feedback. I've added corrections and hopefully clarified the nested fields. We can remove the matching of nested fields and match at the top level only, that would simplify.

I've explained the nested fields through examples, so maybe we can add the examples below to the READ.me ? Or we can remove altogether. I think it's not too complex, but then again I implemented this. I'll let you decode :-)

lib/dfe/analytics/event_matcher.rb
duncanjbrown: Lovely, thanks

stevenleggdfe (commented):
Looks good to me; I think @duncanjbrown should approve though rather than me.

stevenleggdfe (approved):

PR #65, opened 2023-02-23
Update advice on queue config and running imports

asatwal:
[Trello-887](https://trello.com/c/lUFIEQhg/887-add-more-advice-to-dfe-analytics-documentation)

Add following advice to README.md:

- Don't run a backfill when you've got a lot of traffic
- Make sure you have a dedicated worker for BQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,13 +170,15 @@ Events are sent to BigQuery by your application's queueing backed via its Active
 
 Events are generated on each web request and database insert/update/delete query. Depending on the architecture of your application, potentially many jobs could be enqueued as users interact with your application.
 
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -12,7 +12,7 @@ en:
           default: true
         queue:
           description: Which ActiveJob queue to put events on
-          default: ":default"
+          default: ":dfe_analytics"
```
duncanjbrown, line 15: This is now at odds with https://github.com/DFE-Digital/dfe-analytics/blob/86ebf54c11da3473a632bf299ed0418ef24b27f1/lib/dfe/analytics.rb#L79

asatwal:
> Good idea, thanks — imo the default behaviour is fine because it will work out of the box on Sidekiq? Fully support adding to the README tho

Yes, good point to have it working OOB. Restored actual config and left advice the README.

PR #71, opened 2023-04-26
v1.9.0

asatwal:
Fixes: Refer serious misconduct / Apply for QTS unable to upgrade to DfE Analytics GEM (v1.8.0)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,20 @@
 # Changelog
 
-## [v1.8.0](https://github.com/DFE-Digital/dfe-analytics/tree/v1.8.0) (2023-04-04)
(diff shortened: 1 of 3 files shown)


duncanjbrown:
Is this bugfix worth a bump all the way to 1.9? could it be 1.8.1?

asatwal:
> Is this bugfix worth a bump all the way to 1.9? could it be 1.8.1?

Only reason for the bump to 1.9.0, was that it's a fairly major bug that breaks the Docker precompile on multiple projects and hence prevents deployment.

I can change to 1.8.1 if that is more in keeping with release version semantics.. 

Also this has been tested now by @vassyz on the refer serious misconduct team, and the deploy works OK.

asatwal:
> Even tho it's a bad bug, I think this is still a patch

Cool. Easier to create a new PR, and close this one..

PR #74, opened 2023-05-04
Minor wording fixes

scruti:
When going through the gem's codebase, I noticed a couple of places where the wording was incorrect or needed to be upgraded:

- Wrong task description matching another rake task.
- A command message displaying a deprecated interface call for a rake task.


duncanjbrown (approved):
👌 thank you!

PR #75, opened 2023-05-18
Send config events

asatwal:
[Trello-1168](https://trello.com/c/KKNNGzR9/1168-dfe-analytics-gem-send-system-config-event)

Send dfe analytics version and required config items from DfE Analytics to BigQuery as an "analytics_initialise" event.

The "analytics_initialise" is a one-off on startup event.

Also, some miscellaneous READ.me updates:

- Document the anonymise data method available
- A further information and instructions on pushing tags

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,18 @@ DfE::Analytics.config.anonymise_web_request_user_id = false
 
 Anonymisation of `user_id` would be required if the source field in the schema is in `analytics_pii.yml` so that analysts can join the IDs together. If the `user_id` is not in `analytics_pii.yml` but is in `analytics.yml` then `user_id` anonymisation would *not* be required so that the IDs could still be joined together.
 
+### Data Anonymisation Algorithm
+
+Generally all PII data should be anonymised, including data that directly or indirect references PII, for example database IDs.
+
+The `dfe-analytics` gem also anonymises such data, if it is configured to do so. If you are anonymising database IDs in your code (in custom events for example), then you should uses the same hashing algorithm for anonymisation that the gem uses.
+
+The following method should be used in your code for anonymisation:
+
+```ruby
+DfE::Analytics.anonymise(value)
+```
+
 ### Adding specs
 
 #### Testing modes
@@ -432,8 +444,11 @@ Please note that page caching is project specific and each project must carefull
 > 	It could be nice to have tests to prove that connectivity to GCP still works after an update, but we aren't setup for that yet.
 3. (Optional) Verify committed `CHANGELOG.md` changes and alter if necessary: `git show`
 4. Push the branch: `git push origin v${NEW_VERSION}-release`, e.g. `git push origin v1.3.0-release`
-5. Push the tags: `git push --tags`
-6. Cut a PR on GitHub with the label `version-release`, and merge once approved
+5. Cut a PR on GitHub with the label `version-release`, and wait for approval 
+6. Once the PR is approved push the tags, immediately prior to merging: `git push --tags`
+7. Merge the PR.
+
+IMPORTANT:  Pushing the tags will immediately make the release available even on a unmerged branch. Therefore, push the tags to Github only when the PR is approved and immediately prior to merging the PR.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 **👉 Send every web request and database update to BigQuery**
 
-**✋ Skip or anonymise fields containing PII**
+**✋ Skip or pseudonymise fields containing PII**. For an explanation of pseudonymisation, see [ICO Guidance](https://ico.org.uk/media/about-the-ico/consultations/4019579/chapter-3-anonymisation-guidance.pdf)
 
 **✌️  Configure and forget**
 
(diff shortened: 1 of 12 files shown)


lib/dfe/analytics/event.rb
```diff
@@ -5,7 +5,14 @@
...
   module Analytics
     class Event
-      EVENT_TYPES = %w[web_request create_entity update_entity delete_entity import_entity].freeze
+      EVENT_TYPES = %w[
+        web_request
+        create_entity
+        update_entity
+        delete_entity
+        import_entity
+        analytics_initialise
```
stevenleggdfe, line 14: @asatwal maybe ignore this (!) but given we've used verb_noun form for the entity event types what would you think of using ```initialise_analytics``` for this new event type for consistency?

README.md
```diff
@@ -260,6 +260,18 @@ DfE::Analytics.config.anonymise_web_request_user_id = false
 
 Anonymisation of `user_id` would be required if the source field in the schema is in `analytics_pii.yml` so that analysts can join the IDs together. If the `user_id` is not in `analytics_pii.yml` but is in `analytics.yml` then `user_id` anonymisation would *not* be required so that the IDs could still be joined together.
 
+### Data Anonymisation Algorithm
```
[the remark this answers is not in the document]
duncanjbrown: TIL!

README.md
```diff
@@ -260,6 +260,18 @@ DfE::Analytics.config.anonymise_web_request_user_id = false
...
+Generally all PII data should be anonymised, including data that directly or indirect references PII, for example database IDs.
+
+The `dfe-analytics` gem also anonymises such data, if it is configured to do so. If you are anonymising database IDs in your code (in custom events for example), then you should uses the same hashing algorithm for anonymisation that the gem uses.
+
+The following method should be used in your code for anonymisation:
+
+```ruby
+DfE::Analytics.anonymise(value)
+```
+
```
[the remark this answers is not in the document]
duncanjbrown: I think that could usefully go into a comment in the `anonymise` method? I like that callers just use `anonymise` and don't have to worry about the algorithm. Not that we're likely to change it, but it feels hygienic.

README.md
```diff
@@ -432,8 +444,11 @@ Please note that page caching is project specific and each project must carefull
 > 	It could be nice to have tests to prove that connectivity to GCP still works after an update, but we aren't setup for that yet.
 3. (Optional) Verify committed `CHANGELOG.md` changes and alter if necessary: `git show`
 4. Push the branch: `git push origin v${NEW_VERSION}-release`, e.g. `git push origin v1.3.0-release`
-5. Push the tags: `git push --tags`
-6. Cut a PR on GitHub with the label `version-release`, and merge once approved
+5. Cut a PR on GitHub with the label `version-release`, and wait for approval 
+6. Once the PR is approved push the tags, immediately prior to merging: `git push --tags`
```
duncanjbrown, line 448: Thank you — that was bugging me!

lib/dfe/analytics.rb
```diff
@@ -103,6 +104,9 @@ def self.initialize!
           model.include(DfE::Analytics::Entities)
         end
       end
+
+      DfE::Analytics::Initialise.trigger_initialise_event
```
duncanjbrown, line 108: We already have an init routine in `railtie.rb` — could we use that?

lib/dfe/analytics/event.rb
duncanjbrown: +1!

lib/dfe/analytics/initialise.rb
```diff
@@ -0,0 +1,34 @@
...
+        initialise_event = DfE::Analytics::Event.new
+                                              .with_type('analytics_initialise')
+                                              .with_data(initialise_data)
+
+        DfE::Analytics::SendEvents.do([initialise_event.as_json])
+      end
+
+      private
+
+      def initialise_data
```
duncanjbrown, line 24: is this `initialisation_data`?

README.md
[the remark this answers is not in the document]
asatwal: @stevenleggdfe agree that pseudonymisation would be better terminology. However, historically it's been called anonymisation in the gem and there are methods such as `DfE::Analytics.anonymise` that are used externally and config variables that would need to be updated as well as all the other documentation. For consistency we would have to update all these and this would be part of a bigger change with external dependencies, so I don't want to tackle this just yet.

Strictly speaking, this is just a hashing algorithm, so I would say it's more correct to call this "hashing" rather than "anonymisation", but that's a different story.

README.md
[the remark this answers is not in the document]
asatwal: Agrees. I'm trying to keep too much technical details out of the READ.me anyway as it's getting very long.

lib/dfe/analytics.rb
asatwal: In `railtie.rb`, I find the App isn't fully initialised, although it depends on where in railtie the methods are called. I suppose it could be called in the `after_initialize` hook  after the `DfE::Analytics.initialize!`call, after which the config is fully initialised, as we need the config loaded before calling `trigger_initialise_event`.

I'll give that a try.

lib/dfe/analytics/event.rb
asatwal: @stevenleggdfe yes `initialise_analytics` sounds better !

README.md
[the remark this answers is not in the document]
asatwal: @stevenleggdfe I decided to action this review comment in the end as this is a sensitive and important area.

The only item I didn't change is the `anonymised_user_agent_and_ip` field in event data for web requests. I thought this item would require a big migration and also this data doesn't identify a user right ? Only a user session ?

I've kept the `DfE:Analytics.anonymise` for backwards compatibility. We can deprecate in future releases. I believe the only team that use this is tv, although I think it's best to send out deprecation warnings and remove in a more release friendly way.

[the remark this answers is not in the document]
stevenleggdfe: Sounds good to me - thank you!

duncanjbrown (approved):
Thanks @asatwal !

PR #77, opened 2023-06-27
Support single table inheritance

duncanjbrown:
When apps use single-table inheritance, multiple Rails models share the same database table - i.e. they are considered a single 'entity' by DfE Analytics.

This isn't good, because it means that at most one model can be associated with a single table. In turn this means that we will not know to attach callbacks to the other models in the STI heirarchy. In practice, due to the order in which `ActiveRecord::Base.descendants` returns the models, this ends up being the alphabetically first child — not the parent, and not any of the siblings.

~Added DNM so @JR-G can test it out on Apply first!~

Confirmed working on Apply

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -102,6 +102,7 @@ GEM
     backport (1.2.0)
     benchmark (0.2.0)
     builder (3.2.4)
+    coderay (1.1.3)
     concurrent-ruby (1.1.10)
     console (1.15.3)
       fiber-local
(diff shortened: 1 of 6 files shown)


JR-G (approved):

asatwal (approved):
Just a few rubocop failures that need fixing - Otherwise LGTM 👍

duncanjbrown:
Looks like it's broken on Ruby 2.7/Rails 6.1. I'm going to check whether anyone's even on that stack

duncanjbrown:
oh, it's an order-dependent failure due to spec activerecord trickery — will fix

PR #78, opened 2023-06-27
Add better error logging when calling BigQuery insert API on error

asatwal:
Currently when doing a full import on the Teaching Vacancies App, calls to the BigQuery insert API are failing.

There are very few clues in the error logs as to which events are causing the failures. As the full import calls the BigQuery insert API in batches of 500, it's also very difficult to pinpoint the event causing the failure.

This PR adds better error logging when calling the API and logs the events causing the failures also.

### DIFF
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -70,10 +70,10 @@ module DfE
       yield(config)
 
       config.enable_analytics                 ||= proc { true }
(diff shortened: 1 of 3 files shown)


duncanjbrown (approved):
👌

scruti:
Coming late to the party to say thank you for how quickly actioned this :clap:

PR #81, opened 2023-06-29
Do not attempt initialization when the user is changing the database

duncanjbrown:
db:* rake commands are often useful to bring the database into line with what DfE::Analytics expects, so it can be painful not to be able to run them when the config is inconsistent. Allow these commands at all times - full checks will run on app or test boot anyway.

I couldn't think of a reasonable way to test this!

### DIFF
--- a/lib/dfe/analytics/railtie.rb
+++ b/lib/dfe/analytics/railtie.rb
@@ -34,7 +34,15 @@ module DfE
       config.after_initialize do
         # internal gem tests will sometimes suppress this so they can test the
         # init process
-        DfE::Analytics.initialize! unless ENV['SUPPRESS_DFE_ANALYTICS_INIT']
+        if running_db_rake_task? || ENV['SUPPRESS_DFE_ANALYTICS_INIT']
+          puts 'Skipping DfE::Analytics initialization'
+        else
+          DfE::Analytics.initialize!
+        end
+      end
+
+      def running_db_rake_task?
+        defined?(Rake) && Rake.application.top_level_tasks.any? { |t| t.start_with?('db:') }
       end
 
       rake_tasks do


duncanjbrown:
rails uses Rake under the hood, so this works for rails too!

--
Duncan Brown
Teacher Services
Excellent teachers for every child

From: Amar Atwal ***@***.***>
Date: Thursday, 29 June 2023 at 19:14
To: DFE-Digital/dfe-analytics ***@***.***>
Cc: BROWN, Duncan7 ***@***.***>, Author ***@***.***>
Subject: Re: [DFE-Digital/dfe-analytics] Do not attempt initialization when the user is changing the database (PR #81)

@asatwal commented on this pull request.

________________________________

In lib/dfe/analytics/railtie.rb<https://github.com/DFE-Digital/dfe-analytics/pull/81#discussion_r1246946782>:

> @@ -34,7 +34,15 @@ class Railtie < Rails::Railtie

       config.after_initialize do

         # internal gem tests will sometimes suppress this so they can test the

         # init process

-        DfE::Analytics.initialize! unless ENV['SUPPRESS_DFE_ANALYTICS_INIT']

+        if running_db_rake_task? || ENV['SUPPRESS_DFE_ANALYTICS_INIT']

db: commands can also be invoked using rails, so possibly need to cater for rails invocation also.

lib/dfe/analytics/railtie.rb
```diff
@@ -34,7 +34,15 @@ class Railtie < Rails::Railtie
...
-        DfE::Analytics.initialize! unless ENV['SUPPRESS_DFE_ANALYTICS_INIT']
+        if running_db_rake_task? || ENV['SUPPRESS_DFE_ANALYTICS_INIT']
+          puts 'Skipping DfE::Analytics initialization'
+        else
+          DfE::Analytics.initialize!
+        end
+      end
+
+      def running_db_rake_task?
+        defined?(Rake) && Rake.application.top_level_tasks.any? { |t| t.start_with?('db:') }
```
thomasleese, line 45: I was trying to see if there was a way we could look at the namespace of the tasks rather than looking at what it starts with, but couldn't work out a nice way of doing it, so this seems fine.

thomasleese (approved):

lib/dfe/analytics/railtie.rb
duncanjbrown: Yeah, it's a bit ick. But it does seem to work

PR #86, opened 2023-09-26
Prevent duplicate fields in blocklist when using STI

ethax-ross:
When the application has models that are configured with single table inheritance the generated blocklist will contain duplicate field names for the table (once for each model).

This is due to the behaviour of `attribute_names` in Rails; it retrieves the names directly from the table columns and doesn't filter based on the context of the subclass.

To avoid this we need to `uniq` the attribute names generated for the blocklist.

Example blocklist prior to change (assuming a parent/child using STI):

```
{:with_model_candidates_43990_9360=>["id", "first_name", "id", "first_name"]}
```

And after change:

```
{:with_model_candidates_43990_9360=>["id", "first_name"]}
```

### DIFF
--- a/lib/dfe/analytics/fields.rb
+++ b/lib/dfe/analytics/fields.rb
@@ -61,7 +61,7 @@ module DfE
       def self.database
         DfE::Analytics.all_entities_in_application
           .reduce({}) do |list, entity|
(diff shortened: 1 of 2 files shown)


ericaporter (approved):
Looks good 👍

PR #87, opened 2023-09-27
Add log message for entities without primary_key, update README.rb

ericaporter:
`dfe-analytics` does not load entities if they do not have a primary key. This issue was discovered when ECF (CPD) were installing the gem and were unable to load a join table which did not have a primary key.

It was decided to update the error messaging / readme to ask that a primary key to be added to the entity in these situations.

This PR includes these additions (1. update to logging message 2. update to README.rb)

Updates to the loading strategy, to make it more robust, will be considered in future.

### DIFF
--- a/lib/dfe/analytics/load_entities.rb
+++ b/lib/dfe/analytics/load_entities.rb
@@ -17,7 +17,14 @@ module DfE
             next
           end
 
-          unless model.primary_key.to_sym == :id
+          primary_key = model.primary_key
+
+          if primary_key.nil?
+            Rails.logger.info("Not processing #{@entity_name} as it does not have a primary key")
+            next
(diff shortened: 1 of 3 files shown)


lib/dfe/analytics/load_entities.rb
```diff
@@ -17,7 +17,14 @@ def run
             next
           end
 
-          unless model.primary_key.to_sym == :id
+          primary_key = model.primary_key
+
+          if primary_key.nil?
+            Rails.logger.info("Not processing #{@entity_name} as it does not have a primary key")
```
asatwal, line 23: 👍

asatwal (approved):
👍 Looks Good.

PR #88, opened 2023-09-28
v1.11.0 release

ericaporter:
- Add log message for entities without primary_key, update README.rb
- Prevent duplicate fields in blocklist when using STI 
- Nightly checksum and checksum mismatch detection assertion
- Fix single table inheritance deprecation warning bug


asatwal (approved):
👍

PR #89, opened 2023-10-06
Wait polynomially longer rather than exponentially

peteryates:
In Rails 7.1.0 `wait: :exponentially_longer` will actually wait polynomially longer and [has been deprecated](https://github.com/rails/rails/pull/49292).

### DIFF
--- a/lib/dfe/analytics/analytics_job.rb
+++ b/lib/dfe/analytics/analytics_job.rb
@@ -2,7 +2,7 @@ module DfE
   module Analytics
     class AnalyticsJob < ActiveJob::Base
       queue_as { DfE::Analytics.config.queue }
-      retry_on StandardError, wait: :exponentially_longer, attempts: 5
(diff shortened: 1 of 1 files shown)


duncanjbrown:
TIL! Thank you 🙏

duncanjbrown:
Might need to skip for rails 6 if it doesn't have that option?

peteryates:
Oh it's ok, I was doing a quick experiment to see how much of a pain upgrading NPQ and ECF would be 😅  There's no urgency from us here.

vassyz:
Maybe you like my approach? https://github.com/DFE-Digital/dfe-analytics/pull/95
Not sure if there's a better way.

ericaporter:
Closing because we are merging https://github.com/DFE-Digital/dfe-analytics/pull/95 which addresses the issue and also deals with Rails 6 test failures

PR #90, opened 2023-10-06
Add ignored_columns to attributes

ericaporter:
On ECF installation and set up of `dfe-analytics` - it was noticed that not all values of three columns were being sent to dataform (`mentor_profile_id`, `school_cohort_id` and `school_id`). On investigation it was noticed that these values are listed as `ignored_columns` on the children classes in a STI relationship in ECF (but not on the parent).

The updates in this PR are to add hidden_columns to attributes and to allow nil values to be sent to dataform.

### DIFF
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -189,7 +189,7 @@ module DfE
 
     def self.add_ignored_columns_to_attributes(model, attributes)
       ignored_columns = model.class.ignored_columns
-      return attributes unless ignored_columns && !ignored_columns.empty?
+      return attributes if ignored_columns.blank?
 
       nil_values_hash = ignored_columns.to_h { |key| [key, nil] }
       attributes.merge!(nil_values_hash)
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -174,6 +174,7 @@ module DfE
     def self.extract_model_attributes(model, attributes = nil)
       # if no list of attrs specified, consider all attrs belonging to this model
       attributes ||= model.attributes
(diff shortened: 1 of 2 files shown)


spec/dfe/analytics/entities_spec.rb
```diff
@@ -153,9 +200,9 @@
         entity.update(last_name: 'GB')
 
         expect(DfE::Analytics::SendEvents).not_to have_received(:perform_later)
-          .with a_hash_including({
-            'event_type' => 'update_entity'
-          })
+        .with([a_hash_including({
+          'event_type' => 'update_entity'
+        })])
       end
```
ericaporter, line 206: Just housekeeping - these were not evaluating because of missing brackets

lib/dfe/analytics.rb
```diff
@@ -186,6 +187,14 @@ def self.extract_model_attributes(model, attributes = nil)
       allowed_attributes.deep_merge(obfuscated_attributes.transform_values { |value| pseudonymise(value) })
     end
 
+    def self.add_ignored_columns_to_attributes(model, attributes)
+      ignored_columns = model.class.ignored_columns
+      return attributes unless ignored_columns && !ignored_columns.empty?
```
asatwal, line 192: I find it difficult to follow this unless and conditions. Would this be better:

```suggestion
      return attributes if ignored_columns.blank?
```

asatwal (approved):
👍 - Just 1 minor comment.

lib/dfe/analytics.rb
ericaporter: now updated 👍

asatwal: @blank? works on `nil` anyway, so if if there was no check on `ignored_columns` it would still work :-)

PR #91, opened 2023-10-12
v1.11.1

ericaporter:
Fix for ignored_columns to be included in payload. This will unblock ECF and allow them to fully install the analytics gem.


asatwal (approved):
👍

PR #93, opened 2023-10-17
Fix for DfE::Analytics crashing on installation

ericaporter:
Relating to this ticket: https://trello.com/c/ApoCu21a/1381-dfeanalytics-crashes-on-installation There is a bug where the gem crashes on installation 

Investigation: Identified the release, PR and then commit which appears to be causing the problem https://github.com/DFE-Digital/dfe-analytics/pull/37/commits/52486c50b8ea5490b27032f8ea683cbb100dcfa9

It would seem that after adding `def self.initialize!` and `DfE::Analytics::Fields.check!` the generators are not running first and the following error message could be observed: 
Screenshot 2023-10-17 at 12 25 26

@asatwal, I have gone for adding a check in the initialization method, returning unless dfe_analytics.rb exists (actually if we go for this solution, it might be better to have a check for all the files the generator should create including the three .yml files).

I know there are a few ways we could tackle this one but this seemed like the most straightforward - keen to know if you think it will be sufficient.

**Testing the fix:** in ECF, in a demo app I spun up to investigate this issue, in Register and in Apply (all locally)

I uninstalled and installed the gem with my fix (pointing to my branch)

Interestingly the fix did not work in either Apply or Register, but that is because of some legacy config in application.rb (config.analytics = config_for(:analytics)) which predates the gem.

I don’t think this matters, it is only worth noting that if there is a service that has this config that tries to install the gem, it might still crash on installation, but removing this config would fix the problem.

Might be worth making apply and register aware of the legacy config so they can remove it.


asatwal (dismissed):
👍 This seems like the simplest check we can do and it fixes the issue and it is sufficient to check just one of the generated files.

Nice work !

lib/dfe/analytics.rb
```diff
@@ -95,6 +95,11 @@ def self.initialize!
         return
       end
 
+      unless File.exist?(Rails.root.join('config/initializers/dfe_analytics.rb'))
+        Rails.logger.info('Warning: DfE Analytics is not set up. Run `bundle exec rails generate dfe:analytics:install')
+        return
+      end
```
asatwal, line 101: Just had a second thought on this. Is this required for Production ? And should we write to the terminal ?

```suggestion
      if !Rails.env.production? && !File.exist?(Rails.root.join('config/initializers/dfe_analytics.rb'))
        message = "Warning: DfE Analytics is not set up. Run: 'bundle exec rails generate dfe:analytics:install'"
        Rails.logger.info(message)
        puts message
        return
      end
```

ericaporter: Yes, good plan!

lib/dfe/analytics.rb
```diff
@@ -95,8 +95,10 @@ def self.initialize!
...
       end
 
-      unless File.exist?(Rails.root.join('config/initializers/dfe_analytics.rb'))
-        Rails.logger.info('Warning: DfE Analytics is not set up. Run `bundle exec rails generate dfe:analytics:install')
+      unless Rails.env.production? || File.exist?(Rails.root.join('config/initializers/dfe_analytics.rb'))
+        message = "Warning: DfE Analytics is not set up. Run: 'bundle exec rails generate dfe:analytics:install'"
+        Rails.logger.info(message)
+        puts message
         return
       end
```
ericaporter, line 103: For my brain, `unless ||` is easier to read but happy to update to `if !`

asatwal (approved):
👍

lib/dfe/analytics.rb
vassyz: Randomly saw this. First time I see `unless ||` described as easier to read 😅. I personally avoid it.

ericaporter: Ha, well, imo it follows the pattern with the other unless just above it, plus I think easier than if doesn't && doesn't @vassyz

PR #94, opened 2023-10-26
Update scheduling for Entity Table Check Job

ericaporter:
The previous iteration of this was low maintenance self-scheduling but when tested on Register, it was not scheduling properly. 

This iteration is for services who have Sidekiq and Sidekiq cron already installed (or who want to install them). 

Syntax is updated to work with Sidekiq and README.rb is updated to give instructions on how to set up the job.

Services will have more control over scheduling and the job will be more reliable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,7 +324,7 @@ which means they can't be joined up.
 
 ### Entity Table Check Job
 
-If you have *Sidekiq* and *Sidekiq-Cron* installed, you might want to enable the *Entity Table Check Job* to verify that the latest version of an entity table in Big Query matches the database. Ideally the job would be scheduled to run every night.
+If you are using a background processing tool or scheduler (such as Sidekiq, Sidekiq-Cron, Resque, Delayed Job or other alternatives), you may want to configure the Entity Table Check Job. This job is designed to ensure the latest version of an entity table in BigQuery is in sync with the database. It is advisable to schedule this job to run on a nightly basis for consistent data verification.
 
 To enable the Entity Table Check Job, update the configuration option in `config/initializers/dfe_analytics.rb`:
 
@@ -332,7 +332,7 @@ To enable the Entity Table Check Job, update the configuration option in `config
 config.entity_table_checks_enabled = true
 ```
 
-Once enabled, you can set up the Sidekiq-Cron job using the following configuration:
+Once enabled, you will need to configure the job according to the syntax and settings of your chosen background processor or scheduler. Below is an example using Sidekiq-Cron, but similar settings apply for other systems like Resque-Scheduler or Delayed Job's recurring jobs:
 
 ```
 entity_table_check_job:
--- a/README.md
+++ b/README.md
@@ -336,7 +336,7 @@ Once enabled, you can set up the Sidekiq-Cron job using the following configurat
 
 ```
 entity_table_check_job:
-  cron: "0 0 * * *", #Every day at midnight
+  cron: "0 0 * * *", #e.g. every day at midnight
   class: "DfE::Analytics::EntityTableCheckJob",
   queue: default
 ```
--- a/README.md
+++ b/README.md
@@ -324,11 +324,21 @@ which means they can't be joined up.
 
 ### Entity Table Check Job
 
-The entity table check job will run every night and sends data to verify that the latest version of an entity table in BigQuery matches the database. The job is defaulted to false but can be changed by updating the configuration option in
-`config/initializers/dfe_analytics.rb`:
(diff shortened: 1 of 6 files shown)


asatwal (changes requested):
@ericaporter - A few comments for your perusal.

Also we need to change the way we calculate the checksum. The checksum needs to be done on the postgres server rather than it being done on the rails server. I've updated this trello ticket [here](https://trello.com/c/vUVoiYfJ/1510-entity-table-check-job-cheksum-updates-update-to-use-sql) with the details. I expect we'll do that as a seperate PR ?

ericaporter:
> @ericaporter - A few comments for your perusal.
> 
> Also we need to change the way we calculate the checksum. The checksum needs to be done on the postgres server rather than it being done on the rails server. I've updated this trello ticket [here](https://trello.com/c/vUVoiYfJ/1510-entity-table-check-job-cheksum-updates-update-to-use-sql) with the details. I expect we'll do that as a seperate PR ?

@asatwal Thanks for reviewing! Will take a look at those requested updates. Separate PR for the SQL update was my intention, wondering if we could deploy them in succession and do a bit of benchmarking?

lib/dfe/analytics/initialisation_events.rb
```diff
@@ -36,7 +36,7 @@ def send_initialisation_events
           DfE::Analytics::SendEvents.perform_now([initialise_analytics_event])
         end
 
-        DfE::Analytics::EntityTableCheckJob.perform_later if DfE::Analytics.entity_table_checks_enabled?
+        DfE::Analytics::EntityTableCheckJob.perform_async if defined?(Sidekiq) && DfE::Analytics.entity_table_checks_enabled?
```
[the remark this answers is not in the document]
ericaporter: 👍 now updated

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -4,9 +4,10 @@
...
-    # Reschedules to run every 24hours
-    class EntityTableCheckJob < AnalyticsJob
-      WAIT_TIME = Date.tomorrow.midnight
+    # Reschedules with sidekiq_cron to run every 24hours
+    class EntityTableCheckJob
+      include Sidekiq::Worker if defined?(Sidekiq::Worker)
+
       TIME_ZONE = 'London'
 
       def perform
```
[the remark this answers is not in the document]
ericaporter: 👍 now updated

README.md
```diff
@@ -324,13 +324,23 @@ which means they can't be joined up.
 
 ### Entity Table Check Job
 
-The entity table check job will run every night and sends data to verify that the latest version of an entity table in BigQuery matches the database. The job is defaulted to false but can be changed by updating the configuration option in
-`config/initializers/dfe_analytics.rb`:
+If you have *Sidekiq* and *Sidekiq-Cron* installed, you might want to enable the *Entity Table Check Job* to verify that the latest version of an entity table in Big Query matches the database. Ideally the job would be scheduled to run every night.
```
asatwal, line 327: Question here. Do we have to limit this to SideKiq ? I expect this would still work if they use Rescue or Delayed job for example. It's just we've passed on the responsibility of scheduling to the application.

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -4,25 +4,26 @@
 
 module DfE
   module Analytics
-    # Reschedules to run every 24hours
-    class EntityTableCheckJob < AnalyticsJob
-      WAIT_TIME = Date.tomorrow.midnight
+    # Reschedules with sidekiq_cron to run every 24hours
+    class EntityTableCheckJob
```
asatwal, line 8: Analytics job contains the DfE analytics queue as well as the retry config, could we not keep this.

Also my knowledge on this isn't great, but we we necessarily need to include `Sidekiq::Worker` ? Is it required for cron scheduling ?

ericaporter: My initial thinking was that we would need to include it if services do not have Sidekiq installed as an active record adapter, but from what I can see, this is now the standard Sidekiq setup and a check of the services - they all have it installed like this so I can remove the include and also revert to inheriting from AnalyticsJob

README.md
ericaporter: No we don't need to limit it, it would still work if we revert to using `perform_later` within the job rather than using `perform_async` - depending on how Rescue or Delayed Job is set up they might have to change the `perform_later` - but perform_later I think would still be ok without being updated if Sidekiq is configured as the ActiveJob queue adapter

asatwal (approved):
👍 Nice work !

PR #96, opened 2023-11-02
Update lookup to SQL if postgres db

ericaporter:
If the production database is Postgres, to speed up the query, the entity table data / checksum lookup is updated to be entirely in SQL so that it is executed within the db. This will reduce the memory burden on Rails and also prevent the retrieval of a large number of IDs. 

It was necessary to keep the original Active Record lookup given that the test db is SQLite3 and the updated SQL is incompatible.

### DIFF
--- a/lib/dfe/analytics/entity_table_check_job.rb
+++ b/lib/dfe/analytics/entity_table_check_job.rb
@@ -35,11 +35,11 @@ module DfE
         checksum_calculated_at = fetch_current_timestamp_in_time_zone
 
         row_count, checksum = fetch_checksum_data(model, adapter_name, checksum_calculated_at)
-        {
-          row_count: row_count,
-          checksum: checksum,
-          checksum_calculated_at: checksum_calculated_at
-        }
+          { 
+            row_count: row_count, 
+            checksum: checksum, 
+            checksum_calculated_at: checksum_calculated_at 
+          }
       end
 
       def supported_adapter_and_environment?(adapter_name)
@@ -52,7 +52,7 @@ module DfE
 
       def fetch_current_timestamp_in_time_zone
         result = ActiveRecord::Base.connection.select_all('SELECT CURRENT_TIMESTAMP AS current_timestamp')
-        result.first['current_timestamp'].in_time_zone(TIME_ZONE)
+        result.first['current_timestamp'].in_time_zone(TIME_ZONE).iso8601(6)
       end
 
       def fetch_checksum_data(model, adapter_name, checksum_calculated_at)
@@ -68,7 +68,7 @@ module DfE
         checksum_calculated_at_sanitized = ActiveRecord::Base.connection.quote(checksum_calculated_at)
         checksum_sql_query = <<-SQL
           SELECT COUNT(*) as row_count,
-            MD5(STRING_AGG(CHECKSUM_TABLE.ID, '' ORDER BY CHECKSUM_TABLE.UPDATED_AT )) as checksum
+            MD5(COALESCE(STRING_AGG(CHECKSUM_TABLE.ID, '' ORDER BY CHECKSUM_TABLE.UPDATED_AT), '')) as checksum
           FROM (
             SELECT #{sanitized_table_name}.id::TEXT as ID,
                    #{sanitized_table_name}.updated_at as UPDATED_AT
--- a/lib/dfe/analytics/entity_table_check_job.rb
+++ b/lib/dfe/analytics/entity_table_check_job.rb
@@ -14,9 +14,9 @@ module DfE
         DfE::Analytics.entities_for_analytics.each do |entity_name|
           DfE::Analytics.models_for_entity(entity_name).each do |model|
             entity_table_check_event = DfE::Analytics::Event.new
-                                                            .with_data(entity_table_check_data(model))
                                                             .with_type('entity_table_check')
                                                             .with_entity_table_name(model.table_name)
+                                                            .with_data(entity_table_check_data(model))
                                                             .as_json
             DfE::Analytics::SendEvents.perform_later([entity_table_check_event])
             Rails.logger.info("Processing data for #{model.table_name} with row count #{model.count}")
@@ -25,36 +25,48 @@ module DfE
       end
 
       def entity_table_check_data(model)
-        checksum_calculated_at = Time.now.in_time_zone(TIME_ZONE).iso8601
+        checksum_calculated_at = Time.now.in_time_zone(TIME_ZONE).iso8601(6)
+        adapter_name = ActiveRecord::Base.connection.adapter_name.downcase
+        row_count, checksum = fetch_checksum_data(model, checksum_calculated_at, adapter_name)
 
-        if ActiveRecord::Base.connection.adapter_name.downcase == 'postgresql'
-          sql_query = <<-SQL
-            SELECT MD5(STRING_AGG(CHECKSUM_TABLE.ID, ''))
-            FROM
-            (SELECT #{model.table_name}.id::TEXT as ID
-            FROM #{model.table_name}
-            WHERE #{model.table_name}.updated_at < ?
-            ORDER BY #{model.table_name}.updated_at ASC) CHECKSUM_TABLE
-          SQL
+        {
+          row_count: row_count,
+          checksum: checksum,
+          checksum_calculated_at: checksum_calculated_at
+        }
+      end
 
-          result = ActiveRecord::Base.connection.execute(sql_query, Time.parse(checksum_calculated_at)).first
-          {
-            row_count: result['row_count'].to_i,
-            checksum: result['checksum'],
-            checksum_calculated_at: checksum_calculated_at
-          }
+      def fetch_checksum_data(model, checksum_calculated_at, adapter_name)
+        if adapter_name == 'postgresql'
+          fetch_postgresql_checksum_data(model, checksum_calculated_at)
         else
-          table_ids =
-            model
-            .where('updated_at < ?', Time.parse(checksum_calculated_at))
-            .order(updated_at: :asc)
-            .pluck(:id)
-          {
-            row_count: table_ids.count,
-            checksum: Digest::SHA256.hexdigest(table_ids.join),
-            checksum_calculated_at: checksum_calculated_at
-          }
+          fetch_generic_checksum_data(model, checksum_calculated_at)
         end
+      end
+
+      def fetch_postgresql_checksum_data(model, checksum_calculated_at)
+        checksum_sql_query = <<-SQL
+          SELECT COUNT(*) as row_count,
+            MD5(STRING_AGG(CHECKSUM_TABLE.ID, '')) as checksum
+          FROM (
+            SELECT #{model.table_name}.id::TEXT as ID
+            FROM #{model.table_name}
+            WHERE #{model.table_name}.updated_at < '#{checksum_calculated_at}'
+            ORDER BY #{model.table_name}.updated_at ASC
+          ) CHECKSUM_TABLE
+        SQL
+
+        result = ActiveRecord::Base.connection.execute(checksum_sql_query).first
+        [result['row_count'].to_i, result['checksum']]
+      end
+
+      def fetch_generic_checksum_data(model, checksum_calculated_at)
+        table_ids =
+          model
+          .where('updated_at < ?', Time.parse(checksum_calculated_at))
+          .order(updated_at: :asc)
+          .pluck(:id)
+        [table_ids.count, Digest::SHA256.hexdigest(table_ids.join)]
       end
     end
   end
--- a/lib/dfe/analytics/entity_table_check_job.rb
+++ b/lib/dfe/analytics/entity_table_check_job.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'active_support/values/time_zone'
-require 'pry'
 
 module DfE
   module Analytics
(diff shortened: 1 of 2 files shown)


lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -27,16 +27,47 @@ def perform
...
           model
           .where('updated_at < ?', Time.parse(checksum_calculated_at))
           .order(updated_at: :asc)
           .pluck(:id)
-        {
-          row_count: table_ids.count,
-          checksum: Digest::SHA256.hexdigest(table_ids.join),
-          checksum_calculated_at: checksum_calculated_at
-        }
+        [table_ids.count, Digest::SHA256.hexdigest(table_ids.join)]
```
asatwal, line 70: This should be MD5 now rather than SHA256 to keep both versions consistent. I believe the method is: Digest::MD5.

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -27,16 +27,47 @@ def perform
...
       end
 
       def reschedule_job
```
asatwal, line 73: We can remove the `reschedule_job` right ?

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -27,16 +27,47 @@ def perform
 
       def entity_table_check_data(model)
         checksum_calculated_at = Time.now.in_time_zone(TIME_ZONE).iso8601(6)
```
asatwal, line 29: Just had a thought on `calculated_at`. Should we get the current timestamp from the database as the server timestamp and postgres timestamp could be out of sync ?

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -27,16 +27,47 @@ def perform
...
+        adapter_name = ActiveRecord::Base.connection.adapter_name.downcase
+        row_count, checksum = fetch_checksum_data(model, checksum_calculated_at, adapter_name)
+
```
asatwal, line 32: Should be say:
``` Ruby
if adapter_name == 'postgresql' && Rails.env.production?
   Rails.logger.info('Only Postgres databases supported on PRODUCTION')
   return
end 
```

asatwal (commented):
A few more comments for your perusal. I think it's a good idea to merge this PR before the scheduling one.

lib/dfe/analytics/entity_table_check_job.rb
ericaporter: So that is removed in the other PR , I can remove it here if we deploy this first if that is what you mean?

lib/dfe/analytics/entity_table_check_job.rb
ericaporter: @asatwal I'm not sure I follow the logic here - would we want the log message in this scenario or only if it is not postgres and production?

asatwal: @ericaporter  - Apologies I got that wrong.. It's NOT Postgres - Corrected above.

lib/dfe/analytics/entity_table_check_job.rb
asatwal: Ops. Sorry getting the 2 PRs mixed up. The other PR is the correct one to remove this :-)

lib/dfe/analytics/entity_table_check_job.rb
ericaporter: 👍 Good plan, now updated

ericaporter: 👍  now updated

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -26,17 +26,54 @@ def perform
...
+        else
+          fetch_generic_checksum_data(model)
+        end
+      end
+
+      def fetch_postgresql_checksum_data(model)
+        checksum_sql_query = <<-SQL
+          SELECT COUNT(*) as row_count,
+            MD5(STRING_AGG(CHECKSUM_TABLE.ID, '')) as checksum,
+            CURRENT_TIMESTAMP(6) as checksum_calculated_at
```
asatwal, line 56: Will the `CURRENT_TIMESTAMP(6)` here be the same as `CURRENT_TIMESTAMP(6)` in the where clause. I thought it possibly could be different if the query takes a long time to run.

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -26,17 +26,54 @@ def perform
       end
 
       def entity_table_check_data(model)
+        adapter_name = ActiveRecord::Base.connection.adapter_name.downcase
+
+        if adapter_name != 'postgresql' && Rails.env.production?
+          Rails.logger.info('DfE::Analytics: Entity checksum: Only Postgres databases supported on PRODUCTION')
+          return
+        end
+
```
asatwal, line 35: Might be safer to get the current timestamp like this:

``` Ruby
checksum_calculated_at = 
  ActiveRecord::Base.connection.execute('SELECT CURRENT_TIMESTAMP(6)')[0]['current_timestamp']
```

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -26,17 +26,54 @@ def perform
...
+            WHERE #{model.table_name}.updated_at < CURRENT_TIMESTAMP(6)
+            ORDER BY #{model.table_name}.updated_at ASC
+          ) CHECKSUM_TABLE
+        SQL
+
+        result = ActiveRecord::Base.connection.execute(checksum_sql_query).first
+        [result['row_count'].to_i, result['checksum'], result['checksum_calculated_at']]
+      end
+
+      def fetch_generic_checksum_data(model)
```
asatwal, line 69: current_timestamp could be passed into this method rather than calculating. Please see comment above.

lib/dfe/analytics/entity_table_check_job.rb
ericaporter: @asatwal - My thought process was that this SELECT would be transactional and so the timestamp would be the same throughout, but your suggestion might be a better way. Maybe @stevenleggdfe might have some insight?

stevenleggdfe: @asatwal @ericaporter what would you think about ```MAX(updated_at) AS checksum_calculated_at``` instead? Or would that hit performance too heavily?

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -26,17 +26,54 @@ def perform
...
+      def fetch_postgresql_checksum_data(model)
+        checksum_sql_query = <<-SQL
+          SELECT COUNT(*) as row_count,
+            MD5(STRING_AGG(CHECKSUM_TABLE.ID, '')) as checksum,
+            CURRENT_TIMESTAMP(6) as checksum_calculated_at
+          FROM (
+            SELECT #{model.table_name}.id::TEXT as ID
+            FROM #{model.table_name}
+            WHERE #{model.table_name}.updated_at < CURRENT_TIMESTAMP(6)
+            ORDER BY #{model.table_name}.updated_at ASC
```
stevenleggdfe, line 61: @ericaporter suggest omitting the ORDER BY - I'd assume this adds additional processing that we don't use?

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -26,17 +26,54 @@ def perform
...
+          fetch_postgresql_checksum_data(model)
+        else
+          fetch_generic_checksum_data(model)
+        end
+      end
+
+      def fetch_postgresql_checksum_data(model)
+        checksum_sql_query = <<-SQL
+          SELECT COUNT(*) as row_count,
+            MD5(STRING_AGG(CHECKSUM_TABLE.ID, '')) as checksum,
```
stevenleggdfe, line 55: @ericaporter I think this is missing an ORDER BY inside the STRING_AGG() - not in the child query (see comment below)

stevenleggdfe (changes requested):
See comments

lib/dfe/analytics/entity_table_check_job.rb
ericaporter: It looks like `MAX(updated_at)` would be much more accurate than getting the timestamp at the beginning of the calculation (my suggestion), but I guess could potentially cause performance issues. I am wondering if @asatwal's suggestion might be somewhere in between these two where trade offs are concerned?

lib/dfe/analytics/entity_table_check_job.rb
ericaporter: @stevenleggdfe Ah good to know! So omit the subquery ORDER and update the STRING_AGG to be something like`MD5(STRING_AGG(CHECKSUM_TABLE.ID, '' ORDER BY CHECKSUM_TABLE.UPDATED_AT))` ?

stevenleggdfe: Yes, assuming that works in Postgres. That would be how I'd do it in GoogleSQL.

lib/dfe/analytics/entity_table_check_job.rb
asatwal: updated_at is not an indexed column on most database tables, so this may well cause performance problems on large tables. I think we need to avoid any chance of performance problems after our last experience.

I don't believe it will make much of difference to accuracy as once we have a timestamp and any transactions updated after this timestamp will be excluded, both in our calculations and the BQ End.  Can't think of a scenario where MAX(UPDATED_AT) would be more accurate, unless I'm not thinking hard enough :-)

stevenleggdfe: I guess I'm thinking real edge cases like if some update is processing in parallel and hasn't been written to the db yet. Not sure if that's even possible though or whether that would cause a problem anyway. So if it creates performance issues let's just stick with current_timestamp(6) I think.

asatwal: > I guess I'm thinking real edge cases like if some update is processing in parallel and hasn't been written to the db yet. Not sure if that's even possible though or whether that would cause a problem anyway. So if it creates performance issues let's just stick with current_timestamp(6) I think.

Indeed, maybe a possibility that you've taken the `current_timestamp` and a long parallel process that was running before you took the timestamp finishes and writes to the database just as you're calculating the checksum. Even in that case the `updated_at` should be slightly later than your `current_timestamp` assuming that `updated_at` is calculated on the write to the database and not before. I think it's a risk we can take for now.

lib/dfe/analytics/entity_table_check_job.rb
ericaporter: @asatwal I had to adapt this a little to make it compatible with SQLite3 as well as Postgres, kept the concept though

lib/dfe/analytics/entity_table_check_job.rb
ericaporter: Now passing in the current_timestamp to both fetches

stevenleggdfe (commented):
No further comments from but not mine to approve :-)

lib/dfe/analytics/entity_table_check_job.rb
```diff
@@ -25,18 +21,74 @@ def perform
         reschedule_job
       end
 
+      def build_event_for(model)
+        DfE::Analytics::Event.new
```
asatwal, line 25: @ericaporter @stevenleggdfe  - If the row_count is zero ie the table is empty, do we want to handle that situation ? Either not send the event at all OR make sure the code doesn't error both in dfe-analytics and in dataform.

ericaporter: @stevenleggdfe what should I avoid for dataform?

stevenleggdfe: Good question @asatwal .

My preferred approach would be that if a table is empty we *do* send a checksum event to BigQuery with a ```row_count``` of ```0``` and a ```NULL``` value (not an empty ```value``` array - BQ can't store that - or an empty string as ```MD5("")``` produces a non null or empty value in GoogleSQL) for the checksum. Is that feasible?

stevenleggdfe: Note in Google SQL ```SELECT MD5("")``` produces ```1B2M2Y8AsgTpgAmY7PhCfg==``` !

stevenleggdfe: This might be important in cases where all the rows have been deleted in the database but for some reason we still have them in BQ because something went wrong with the deletion events.

asatwal: So for the Postgres SQL if run locally:

```SQL
SELECT count (*)  as row_count, MD5(STRING_AGG(CHECKSUM_TABLE.ID, '' ORDER BY CHECKSUM_TABLE.UPDATED_AT)) as checksum
FROM 
(SELECT trainees.id::TEXT as ID,
        updated_at as UPDATED_AT
 FROM trainees
 WHERE updated_at < '2020-11-05 09:25:19.863415'
 ORDER BY trainees.updated_at ASC) CHECKSUM_TABLE;
```
Result :
row_count  checksum
----------------------
0                  NULL

In Ruby:
Digest::MD5.hexdigest('')
=> "d41d8cd98f00b204e9800998ecf8427e"

In Postgres SQL:
``` SQL
SELECT MD5('');
```
Result:
md5
----------------------------------------
d41d8cd98f00b204e9800998ecf8427e

Which is different from BQ. Which makes me think do we have the same MD5 in Google SQL as Ruby and Postgres

stevenleggdfe: @asatwal good spot. It's not the same it appears.

However I think the difference is that we're missing some conversion to/from hexadecimal.

In GoogleSQL ```SELECT TO_HEX(MD5(""))``` produces ```d41d8cd98f00b204e9800998ecf8427e```, which matches.

I'll update the dfe-analytics-dataform logic to match.

ericaporter: @stevenleggdfe I think sending nil values as an empty array is a design choice in dfe-analytics - even if I would try to send a nil value, it will get converted to an `[]`

stevenleggdfe: @ericaporter oh sorry, my bad! ```[]``` is fine. ```[NULL]``` is not. ```[""]``` is not great either.

asatwal (approved):
👍  LGTM

lib/dfe/analytics/entity_table_check_job.rb
ericaporter: I have kept `MD5(COALESCE(STRING_AGG(CHECKSUM_TABLE.ID, '' ORDER BY CHECKSUM_TABLE.UPDATED_AT ASC), ''))` and `Digest::MD5.hexdigest(table_ids.join)` and both will return `[0, “d41d8cd98f00b204e9800998ecf8427e”]`

stevenleggdfe (approved):

PR #99, opened 2023-11-30
Remove fix for ignored columns

ericaporter:
This fix was introduced to assist ECF with a STI bug where different ignored_columns on different child models (with the same parent) were not being sent to BigQuery.

The fix wasn't working all of the time and in the end the simplest solution was to remove reference to the ignored_columns in the child models so that all of the fields are being sent all of the time.

It is probably confusing to keep this code in dfe-analytics and it is unlikely to be needed by another service because usually when a column is ignored there would not be a requirement to send it to Big Query.


asatwal (approved):
👍 LGTM

PR #106, opened 2024-01-03
Ensure order by column is in analytics.yml

ericaporter:
We are now dynamically finding the order by column for the entity since sometimes updated_at is missing.

Instead of only checking the model for the existence of the order_by fields, we are now ensuring that they exist in the analytics.yml.

The logic has been updated so that we are from within the determine_order_column method ->
1) finding the entity keys on the allow list
2) returning unless an order column exists i.e. whether updated_at / created_at / id exists in the analytics.yml
3) adding an additional check that the column exists both on the model and in the analytics.yml


asatwal (approved):
Nice work!

PR #107, opened 2024-01-08
Release v1.11.6

ericaporter:
This update ensures that the checksum lookup only uses order_by values that exist in the analytics.yml (instead of looking to the model for their presence)

It also removes ordering by ID although this may be reintroduced at a later stage.


asatwal (approved):
👍 LGTM

PR #108, opened 2024-01-24
Update import job to send unique timestamp and import_entity_table_check event

ericaporter:
The current DfE Analytics import job (aka Backfill) in ruby imports all BQ entities to BQ. The job needs further enhancements to allow DfE Analytics to do a reliable sync of a complete table ie update to the latest data for each row and delete any  extraneous rows in BQ.

This PR includes the following changes:
1) Adds a single unique import entity id to ALL events for a given import as a numerical timestamp in the format: YYYYMMDDHHMMSS. The field for this in the event could be event_tag

2) For each entity, once the import has been completed, a new checksum event is sent (import_entity_table_check)
which contains the unique entity import id to the event_tag

3) A checksum event can now be ordered by ID if created_at / updated_at are missing. Note that this changed will also affect the nightly checksum job

4) The entity table check and checksum calculation logic has been refactored into service objects


lib/dfe/analytics/entity_processor.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'shared/checksum_logic'
+
+module DfE
+  module Analytics
+    # Processes entities by sending the import_entity_check_event
+    class EntityProcessor
```
asatwal, line 6: The `EntityProcessor` and the `ChecksumLogic` is factoring out the common logic and is a step in the right direction. I think we should go a step further though and create a service object for the table check logic.

Most of the rails projects use the following service pattern:

``` Ruby
module ServicePattern
  def self.included(base)
    base.extend(ClassMethods)
    base.class_eval do
      private_class_method(:new)
    end
  end

  def call
    raise(NotImplementedError("#call must be implemented"))
  end

  module ClassMethods
    def call(*args, **keyword_args, &block)
      return new.call if args.empty? && keyword_args.empty? && block.nil?

      new(*args, **keyword_args, &block).call
    end
  end
end
```

I won't go into the details here as it's too much typing, but we can have a chat about it.

lib/dfe/analytics/shared/checksum_logic.rb
```diff
@@ -0,0 +1,102 @@
...
+      order_column: order_column
+    }
+  end
+
+  def determine_order_column(entity, columns)
+    if ActiveRecord::Base.connection.column_exists?(entity, :updated_at) && columns.include?('updated_at')
+      'UPDATED_AT'
+    elsif ActiveRecord::Base.connection.column_exists?(entity, :created_at) && columns.include?('created_at')
+      'CREATED_AT'
+    else
```
asatwal, line 38: We need to reintroduce the order by "ID" column.

lib/dfe/analytics/load_entity_batch.rb
```diff
@@ -5,33 +5,44 @@ class LoadEntityBatch < AnalyticsJob
...
+        model_class.where(id: ids).map do |record|
+          build_event(record, model_class.table_name, import_entity_id)
         end
       end
+
+      def build_event(record, table_name, import_entity_id)
+        DfE::Analytics::Event.new
+            .with_type('import_entity')
+            .with_entity_table_name(table_name)
+            .with_tags(import_entity_id)
```
stevenleggdfe, line 43: @ericaporter will this be sufficient to allow us to tell the difference between import events from two different imports even if they are for the same entity?

stevenleggdfe: Looking at line 16 of lib/dfe/analytics/tasks/import_entities.rake I think it would be? But if so maybe we could call it something different? I thought this was referring to the UID of an entity, or the name of an entity (e.g. application_choices) rather than the UID of the import. Maybe entity_import_id or something?

lib/dfe/analytics/entity_processor.rb
ericaporter: Fully onboard!

lib/dfe/analytics/shared/checksum_logic.rb
ericaporter: Oh yes, I'd forgotten about that!

lib/dfe/analytics/load_entity_batch.rb
ericaporter: Happy to rename if it makes things easier

asatwal: I suppose the name `import_entity_id` may be confusing, but it doesn't actually get added to the event. Only the value gets added as a tag. The value is a numeric date BTW.

What about `import_entities_job_id` OR just `import_job_id` ?

lib/dfe/analytics/services/entity_table_checks.rb
```diff
@@ -0,0 +1,117 @@
...
+          }
+        end
+
+        def send_entity_table_check_event(entity_name, entity_type, entity_tag, order_column)
+          entity_table_check_event = build_event_for(entity_name, entity_type, entity_tag, order_column)
+          DfE::Analytics::SendEvents.perform_later([entity_table_check_event])
+        end
+
+        def build_event_for(entity_name, entity_type, entity_tag, order_column)
+          unless DfE::Analytics.models_for_entity(entity_name).any?
```
ericaporter, line 102: We could extract an event_builder service

lib/dfe/analytics/services/entity_table_checks.rb
```diff
@@ -0,0 +1,117 @@
...
+          {
+            row_count: row_count,
+            checksum: checksum,
+            checksum_calculated_at: checksum_calculated_at,
+            order_column: order_column
+          }
+        end
+
+        def send_entity_table_check_event(entity_name, entity_type, entity_tag, order_column)
+          entity_table_check_event = build_event_for(entity_name, entity_type, entity_tag, order_column)
```
ericaporter, line 97: We could extract an event_sender service

lib/dfe/analytics/services/entity_table_checks.rb
```diff
@@ -0,0 +1,117 @@
...
+        def supported_adapter_and_environment?
+          return true if %w[postgresql postgis].include?(adapter_name) || !Rails.env.production?
+
+          Rails.logger.info('DfE::Analytics: Entity checksum: Only Postgres databases supported on PRODUCTION')
+
+          false
+        end
+
+        def fetch_current_timestamp_in_time_zone
+          result = connection.select_all('SELECT CURRENT_TIMESTAMP AS current_timestamp')
```
ericaporter, line 48: We could extract a current timezone fetcher

lib/dfe/analytics/services/postgres_checksum_calculator.rb
```diff
@@ -0,0 +1,56 @@
+require_relative '../shared/service_pattern'
```
asatwal, line 1: Maybe a simple collaboration spec to check the connection is executed with the expected SQL statement ?

asatwal (commented):
This is looking really good @ericaporter . Don't think you need to turn any other of the other smaller modules into services. Just a few more simple collaboration specs I think ?

asatwal (approved):
The collaboration specs for the services are a nice to have.

lib/dfe/analytics/services/checksum_calculator.rb
```diff
@@ -0,0 +1,38 @@
+require_relative '../shared/service_pattern'
```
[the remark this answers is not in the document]
ericaporter: Now added

lib/dfe/analytics/services/generic_checksum_calculator.rb
```diff
@@ -0,0 +1,52 @@
+require_relative '../shared/service_pattern'
```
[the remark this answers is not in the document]
ericaporter: Now added 👍

lib/dfe/analytics/services/postgres_checksum_calculator.rb
ericaporter: Now added 👍

asatwal (approved):
This is looking good @ericaporter - With hindsight I can see why PG gets confused by the ID column ordering in the Checksum subquery. Well done for finding the problem and persisting with this.

PR #109, opened 2024-01-26
Update conditional to handle 'postgis' db

ericaporter:
Teaching vacancies use PostGIS as a db which extends the capabilities of the PostgreSQL.

The Entity Table check job is not running for them on their review app as it point to production and the adapter_name is postgis. In production the job also won't run because the conditional returns false unless it is postgres in production.

This PR updates the conditional so that the job will run if the adapter_name is postgis. This shouldn't cause any issues as postgis is an extension of postgres and so the lookup should run as it would on postgres


asatwal (approved):
👍  LGTM!

PR #111, opened 2024-02-01
Enable filtering of request query strings using `ActiveSupport::ParameterFilter`.

steventux:
PII can leak into DfE analytics if present in the request query string, a common case is search parameters using personal details.

This PR adds the ability to enable web request event query filtering using the standard Rails parameter filter mechanism via the `DfE::Analytics.filter_web_request_events` configuration value.

See: https://github.com/DFE-Digital/access-your-teaching-qualifications/commit/5941285add123ffbc8ab8a738305c48aad16e1a2 for a common workaround currently being used in services to mitigate this. 

It would be preferable not to replicate this workaround on a per-service basis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,14 @@ user identifier proc can be defined in `config/initializers/dfe_analytics.rb`:
 DfE::Analytics.config.user_identifier = proc { |user| user&.uid }
 ```
 
+#### Filtering PII from web request query string data
+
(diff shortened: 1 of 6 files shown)


steventux:
[Protect PII in the CBL search URL](https://trello.com/c/rNmbNx7Z/222-protect-pii-in-the-cbl-search-url)

PR #112, opened 2024-02-13
Use original_fullpath for request path

thomasleese:
This changes the event to get the request path from the `original_fullpath` method rather than the `path` method. I'm making this change as the `original_fullpath` represents the path of the page as requested by the user, not necessarily the page being rendered.

In most cases the original path and the path will be the same, but there are some scenarios where it is. Specifically, when rendering a custom error page in Rails which is mounted at `/404` the path ends up being the value `/404` rather than the path which triggered the 404 page.


asatwal (approved):
Looks Good @thomasleese 👍

thomasleese:
Thanks @asatwal - I've merged this in. I don't think it needs to be released any time soon so it can wait until the next release.

PR #115, opened 2024-02-21
Add maintenance window logic

ericaporter:
Relating to this [ticket ](https://trello.com/c/LacYDk5E/1802-dfe-analytics-support-scheduled-downtime-for-bigquery)

We need the ability to do maintenance on on a live streamed BiqQuery dataset, such as register_events_production.

This PR adds a configuration item for a maintenance window.

Redis logic - WIP
- On reaching the start maintenance window start timestamp, DfE Analytics will start queueing items with a schedule time of:
end_timestamp + number of seconds elapsed since start_timestamp
The queued items will end up queued in Redis.
- Once the end of the maintenance window is reached, the queue items will be processed and sent to BigQuery and a First In First Out (FIFO) base.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -70,6 +70,6 @@ en:
         bigquery_maintenance_window:
           description: |
             Schedule a maintenance window during which no events are streamed to BigQuery
-            in the format of '22-01-2024 19:30..22-01-2024 20:30'.
+            in the format of '22-01-2024 19:30..22-01-2024 20:30' (UTC).
           default: ENV['BIGQUERY_MAINTENANCE_WINDOW']
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -67,3 +67,9 @@ en:
             return a boolean indicating whether the page is cached and will
             be served by rack middleware.
           default: proc { |_rack_env| false }
+        bigquery_maintenance_window:
+          description: |
+            Schedule a maintenance window during which no events are streamed to BigQuery
+            in the format of '22-01-2024 19:30..22-01-2024 20:30'.
+          default: ENV['BIGQUERY_MAINTENANCE_WINDOW']
+
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -252,8 +252,11 @@ module DfE
 
       start_str, end_str = config.bigquery_maintenance_window.split('..', 2).map(&:strip)
       begin
-        start_time = DateTime.strptime(start_str, '%d-%m-%Y %H:%M')
-        end_time = DateTime.strptime(end_str, '%d-%m-%Y %H:%M')
+        parsed_start_time = DateTime.strptime(start_str, '%d-%m-%Y %H:%M')
+        parsed_end_time = DateTime.strptime(end_str, '%d-%m-%Y %H:%M')
+
+        start_time = Time.zone.parse(parsed_start_time.to_s)
+        end_time = Time.zone.parse(parsed_end_time.to_s)
 
         if start_time > end_time
           Rails.logger.info('Start time is after end time in maintenance window configuration')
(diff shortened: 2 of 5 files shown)


lib/dfe/analytics.rb
```diff
@@ -244,5 +246,32 @@ def self.rack_page_cached?(rack_env)
...
+        [nil, nil]
+      end
+    end
+
+    def self.within_maintenance_window?
+      start_time, end_time = parse_maintenance_window
+      return false unless start_time && end_time
+
+      DateTime.now.between?(start_time, end_time)
+    end
```
asatwal, line 275: We'll need to add a method like this:

``` Ruby 
def self.next_scheduled_time_after_maintenance_window
  start_time, end_time = parse_maintenance_window

  end_time + (Time.zone.now - start_time).seconds
end
```

config/locales/en.yml
```diff
@@ -67,3 +67,9 @@ en:
             return a boolean indicating whether the page is cached and will
             be served by rack middleware.
           default: proc { |_rack_env| false }
+        bigquery_maintenance_window:
+          description: |
+            Schedule a maintenance window during which no events are streamed to BigQuery
+            in the format of '22-01-2024 19:30..22-01-2024 20:30'.
```
asatwal, line 73: May need to specify Timezone in UTC here.

lib/dfe/analytics/send_events.rb
```diff
@@ -18,6 +18,12 @@ def self.do(events)
       end
 
       def perform(events)
+        if DfE::Analytics.within_maintenance_window?
+          # Delaying / Queueing events - the redis bit
+          Rails.logger.info('Within BigQuery maintenance window. Events will be queued for later processing.')
+          return
+        end
+
```
asatwal, line 26: Need to remove this block here and move into the `do` method above.

```suggestion
```

In the `do` method replace lines 13-17 with the following:

``` Ruby
   if DfE::Analytics.within_maintenance_window?
     set(wait_until: DfE::Analytics.next_scheduled_time_after_maintenance_window).perform_later(events)
   elsif DfE::Analytics.async?
     perform_later(events)
   else
     perform_now(events)
   end
 ```

**Review by asatwal — COMMENTED** [review]
See comments in code on suggestions for implementation.

File: lib/dfe/analytics.rb [hunk 4]
```diff
@@ -244,5 +246,32 @@ def self.rack_page_cached?(rack_env)
...
+      start_str, end_str = config.bigquery_maintenance_window.split('..', 2).map(&:strip)
+      begin
+        start_time = DateTime.strptime(start_str, '%d-%m-%Y %H:%M')
+        end_time = DateTime.strptime(end_str, '%d-%m-%Y %H:%M')
+
+        if start_time > end_time
+          Rails.logger.info('Start time is after end time in maintenance window configuration')
+          return [nil, nil]
+        end
+
```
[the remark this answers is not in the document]
ericaporter: Going to hold off on this atm as discussed, since it will be us making these updates, but this might need to be revisited

config/locales/en.yml
ericaporter: Now added to the description

spec/dfe/analytics/send_events_spec.rb
```diff
@@ -163,39 +163,40 @@
...
+        expect(DfE::Analytics::SendEvents).to receive(:set).with(wait_until: expected_wait_until).and_call_original
+        DfE::Analytics::SendEvents.do(events)
+      end
 
-        DfE::Analytics::Testing.webmock! do
-          described_class.new.perform([event.as_json])
+      it 'does not execute the event immediately within the maintenance window' do
+        expect(DfE::Analytics::SendEvents).not_to receive(:perform_now).with(events)
+        DfE::Analytics::SendEvents.do(events)
+      end
```
asatwal, line 193: I think we can also have another test to ensure it doesn't call `perform_later` as the majority of the events are async I believe.

asatwal (approved):
Looks good @ericaporter - Just have a minor comment about adding another test.

spec/dfe/analytics/send_events_spec.rb
ericaporter: Now updated 👍

PR #117, opened 2024-03-04
Remove array brackets for unused event tag

ericaporter:
An error was observed on ECF as the event tag is nil in the context of a create event. Removing the array brackets from the event tag in the entities file should prevent this - although a longer term solution will need to be found.


asatwal (approved):
👍

PR #120, opened 2024-03-18
Add support for workload identity federation and consolidate GCP/BigQuery APIs into seperate classes

asatwal:
Use workload identity federation as an alternative authentication for GCP.

For more details please see GCP documentation [here](https://cloud.google.com/iam/docs/workload-identity-federation). 

Configuration of workload identity federation with  Azure and links to the azure documentation can be found [here](https://cloud.google.com/iam/docs/workload-identity-federation-with-other-clouds#azure)

The steps involved for authentication are shown in the diagram below:

 
azure-gcp-wif

From the diagram, Step 4 is what we have to today with the API key - We go straight to step 4 with the API Key.
So with WIF weeliminate the API Key, but go through Steps 1-3 to get an service account impersonation access token -  This replaces the API Key.

This file:
[lib/dfe/analytics/azure_federated_auth.rb](https://github.com/DFE-Digital/dfe-analytics/pull/120/files#diff-dcb05a6b0fc256ae3ee7b72cd9071ea14cddd36903e3449b58803ae1af554b90)
implements steps 1-3>

Also the streaming API used to send data to BQ need to be changed.
The old API doesn't work with the new Auth and the new Auth doesn't work with the old API, hence the new event client and the legacy event client.
All config need is from either Azure end variables setup by DevOps:
AZURE_CLIENT_ID
AZURE_FEDERATED_TOKEN_FILE
The rest of the config is from the ENV variable:
GOOGLE_CLOUD_CREDENTIALS
This comes from Google and is a downloadable cloud config.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -75,7 +75,8 @@ en:
         azure_federated_auth:
           description: |
             Whether to use azure workload identity federation for authentication
-            instead of the BigQuery API JSON Key
+            instead of the BigQuery API JSON Key. Note that this also will also
+            use a new version of the BigQuery streaming APIs.
           default: false
         azure_client_id:
           description: |
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -60,7 +60,7 @@ en:
         entity_table_checks_enabled:
           description: |
             Whether to run entity table checksum job.
-          default: false    
+          default: false
         rack_page_cached:
           description: |
             A proc which will be called with the rack env, and which should
@@ -72,4 +72,28 @@ en:
             Schedule a maintenance window during which no events are streamed to BigQuery
             in the format of '22-01-2024 19:30..22-01-2024 20:30' (UTC).
           default: ENV['BIGQUERY_MAINTENANCE_WINDOW']
-
+        azure_federated_auth:
+          description: |
+            Whether to use azure workload identity federation for authentication
+            instead of the BigQuery API JSON Key
+          default: false
+        azure_client_id:
+          description: |
+            Client Id of the app in azure
+          default: ENV['AZURE_CLIENT_ID']
+        azure_token_path:
+          description: |
+            Path of token file for used for getting token from azure ad
+          default: ENV['AZURE_FEDERATED_TOKEN_FILE']
+        azure_scope:
+          description: |
+            Azure audience scope
+          default: api://AzureADTokenExchange/.default
+        gcp_scope:
+          description: |
+            Google cloud scope
+          default: https://www.googleapis.com/auth/cloud-platform
+        google_cloud_credentials:
+          description: |
+            Google generated cloud credentials file
+          default: ENV['GOOGLE_CLOUD_CREDENTIALS']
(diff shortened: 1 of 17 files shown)


spec/dfe/analytics/big_query_api_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  describe '#events_client' do
+    it 'raises a configuration error on missing config values' do
+      with_analytics_config(bigquery_project_id: nil) do
+        expect { described_class.events_client }.to raise_error(DfE::Analytics::BigQueryApi::ConfigurationError)
+      end
+    end
+
+    context 'when authorization endoint returns OK response' do
+      it 'calls the expected big query apis' do
```
ericaporter, line 35: 'endpoint'

spec/dfe/analytics/big_query_legacy_api_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  describe '#events_client' do
+    it 'raises a configuration error on missing config values' do
+      with_analytics_config(bigquery_project_id: nil) do
+        expect { described_class.events_client }.to raise_error(DfE::Analytics::BigQueryLegacyApi::ConfigurationError)
+      end
+    end
+
+    context 'when authorization endoint returns OK response' do
+      let(:events_client) { double(:events_client) }
```
ericaporter, line 28: endpoint

config/locales/en.yml
```diff
@@ -72,4 +72,28 @@ en:
             Schedule a maintenance window during which no events are streamed to BigQuery
             in the format of '22-01-2024 19:30..22-01-2024 20:30' (UTC).
           default: ENV['BIGQUERY_MAINTENANCE_WINDOW']
-
+        azure_federated_auth:
+          description: |
+            Whether to use azure workload identity federation for authentication
```
stevenleggdfe, line 77: @asatwal does this also switch between APIs? If so could we add that to the description?

lib/dfe/analytics/big_query_api.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+module DfE
+  module Analytics
+    # For use with for workload identity federeation
```
stevenleggdfe, line 5: @asatwal would it be possible for a team to use this API without WIF? How much effort would it be to add that flexibility? In future I can see the old API being decommissioned and some teams needing to move to the new one but not having been able to get things set up in Azure such that they can use WIF.

config/locales/en.yml
asatwal: Yes this does switch APIs as the old APIs don't work with the new WIF Auth. Good point I'll add to this description.

lib/dfe/analytics/big_query_api.rb
ericaporter: @asatwal I was wondering generally if it's a strong recommend that teams migrate to WIF?

asatwal: I think it is possible to use the new streaming APIs with the old authentication, but this would be effort though. I think we could consider this if there is a real need for this. As Erica points out above, we could insist on WIF initially and it's a strong recommendation.

ericaporter (approved):
👍 Looks good!

PR #122, opened 2024-03-21
Update rakefile

ericaporter:
We've identified that the current annotated tagging process in the dfe-analytics release workflow is leading to complications, particularly with Dependabot's handling of our releases. 

This PR modifies our tagging strategy to create lightweight (non-annotated) tags instead. This adjustment ensures each tag is associated with a single commit hash, streamlining our version history and preventing the confusion experienced by Dependabot.


asatwal (approved):
👍 LGTM

PR #126, opened 2024-04-02
Re-add 'as checksum'

ericaporter:
In the latest re-factoring, 'as checksum' was omitted and this stopped checksums from pulling through. This PR fixes that.


asatwal (approved):
👍

PR #128, opened 2024-04-05
Allow hidden data to be sent separately

ericaporter:
[Ticket](https://trello.com/c/IpCySdeH/1879-populate-datahidden-field-in-streamed-events) 

This PR introduces a with_hidden_data event field
Hidden data is kept separate to both allowed_data and allowed_pii

### DIFF
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -211,7 +211,10 @@ module DfE
       hidden_attributes = attributes.slice(*exportable_hidden_pii_attrs&.map(&:to_s))
 
       # Allowed attributes (which currently includes the allowlist_pii) must be kept separate from hidden_attributes
-      [allowed_attributes.deep_merge(obfuscated_attributes), hidden_attributes]
+      model_attributes = {}
+      model_attributes.merge!(data: allowed_attributes.deep_merge(obfuscated_attributes)) if allowed_attributes.any? || obfuscated_attributes.any?
+      model_attributes.merge!(hidden_data: hidden_attributes) if hidden_attributes.any?
+      model_attributes
     end
 
     def self.anonymise(value)
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -192,16 +192,26 @@ module DfE
     def self.extract_model_attributes(model, attributes = nil)
       # if no list of attrs specified, consider all attrs belonging to this model
       attributes ||= model.attributes
-      table_name = model.class.table_name
+      table_name = model.class.table_name.to_sym
 
-      exportable_attrs = allowlist[table_name.to_sym].presence || []
-      pii_attrs = allowlist_pii[table_name.to_sym].presence || []
+      exportable_attrs = (allowlist[table_name].presence || []).map(&:to_sym)
+      hidden_pii_attrs = (hidden_pii[table_name].presence || []).map(&:to_sym)
+      pii_attrs = (allowlist_pii[table_name].presence || []).map(&:to_sym)
+
+      # Validation in fields.rb ensures attributes do not appear on both allowlist_pii and allowlist_hidden_pii
       exportable_pii_attrs = exportable_attrs & pii_attrs
+      exportable_hidden_pii_attrs = exportable_attrs & hidden_pii_attrs
 
-      allowed_attributes = attributes.slice(*exportable_attrs&.map(&:to_s))
-      obfuscated_attributes = attributes.slice(*exportable_pii_attrs&.map(&:to_s))
+      # Exclude both pii and hidden attributes from allowed_attributes
+      allowed_attrs_to_include = exportable_attrs - (exportable_pii_attrs + exportable_hidden_pii_attrs)
 
-      allowed_attributes.deep_merge(obfuscated_attributes.transform_values { |value| pseudonymise(value) })
+      allowed_attributes = attributes.slice(*allowed_attrs_to_include&.map(&:to_s))
+      obfuscated_attributes = attributes.slice(*exportable_pii_attrs.map(&:to_s))
+                .transform_values { |value| pseudonymise(value) }
+      hidden_attributes = attributes.slice(*exportable_hidden_pii_attrs&.map(&:to_s))
+
+      # Allowed attributes (which currently includes the allowlist_pii) must be kept separate from hidden_attributes
+      [allowed_attributes.deep_merge(obfuscated_attributes), hidden_attributes]
     end
 
     def self.anonymise(value)
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -143,6 +143,8 @@ module DfE
 
     def self.hidden_pii
       Rails.application.config_for(:analytics_hidden_pii)
+    rescue RuntimeError
+      { 'shared' => {} }
     end
 
     def self.blocklist
--- a/lib/dfe/analytics/event.rb
+++ b/lib/dfe/analytics/event.rb
@@ -4,6 +4,7 @@ require 'active_support/values/time_zone'
 
 module DfE
   module Analytics
+    # rubocop:disable Metrics/ClassLength
     class Event
       EVENT_TYPES = %w[
         web_request create_entity update_entity delete_entity import_entity initialise_analytics entity_table_check import_entity_table_check
@@ -78,6 +79,12 @@ module DfE
         self
       end
 
+      def with_hidden_data(hash)
+        @event_hash.deep_merge!(hidden_data: hash_to_kv_pairs(hash))
+
+        self
+      end
+
       def with_tags(tags)
         @event_hash[:event_tags] = tags if tags
 
@@ -137,5 +144,6 @@ module DfE
         user_id
       end
     end
+    # rubocop:enable Metrics/ClassLength
   end
 end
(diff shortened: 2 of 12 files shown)


lib/dfe/analytics/event.rb
```diff
@@ -78,6 +79,12 @@ def with_data(hash)
         self
       end
 
+      def with_hidden_data(hash)
+        @event_hash.deep_merge!(hidden_data: hash_to_kv_pairs(hash))
+
```
ericaporter, line 84: This possibly doesn't need to be deep_merge here if the hidden_data we are sending is not going to be nested

lib/dfe/analytics/load_entity_batch.rb
```diff
@@ -4,27 +4,39 @@ class LoadEntityBatch < AnalyticsJob
...
-        events = create_events(model_class, ids, entity_tag)
+        model_class = resolve_model_class(model_class_arg)
 
-        payload_byte_size = events.sum(&:byte_size_in_transit)
+        # Process the entire batch if it's small enough or if splitting is no longer feasible
+        if ids.size <= MIN_BATCH_SIZE
+          process_batch(model_class, ids, entity_tag)
+        else
+          events = create_events(model_class, ids, entity_tag)
+          payload_byte_size = events.sum(&:byte_size_in_transit)
```
ericaporter, line 20: I introduced MIN_BATCH_SIZE here because with the updated data structure, I was getting recursion between ln 16 and 19. Introducing this seemed to fix it

lib/dfe/analytics/event.rb
asatwal: This is OK I believe, as we can't make assumptions about nesting within the hidden data.

lib/dfe/analytics.rb
```diff
@@ -192,16 +192,26 @@ def self.models_for_entity(entity)
...
-      allowed_attributes = attributes.slice(*exportable_attrs&.map(&:to_s))
-      obfuscated_attributes = attributes.slice(*exportable_pii_attrs&.map(&:to_s))
+      allowed_attributes = attributes.slice(*allowed_attrs_to_include&.map(&:to_s))
+      obfuscated_attributes = attributes.slice(*exportable_pii_attrs.map(&:to_s))
+                .transform_values { |value| pseudonymise(value) }
+      hidden_attributes = attributes.slice(*exportable_hidden_pii_attrs&.map(&:to_s))
 
-      allowed_attributes.deep_merge(obfuscated_attributes.transform_values { |value| pseudonymise(value) })
+      # Allowed attributes (which currently includes the allowlist_pii) must be kept separate from hidden_attributes
+      [allowed_attributes.deep_merge(obfuscated_attributes), hidden_attributes]
```
asatwal, line 214: Not sure if I have misunderstood, but I feel we can avoid a lot of changes if we return like:

``` Ruby 
{
  data: allowed_attributes.deep_merge(obfuscated_attributes),
  hidden_data: hidden_attributes
}
```

lib/dfe/analytics/event.rb
```diff
@@ -78,6 +79,12 @@ def with_data(hash)
         self
       end
 
+      def with_hidden_data(hash)
+        @event_hash.deep_merge!(hidden_data: hash_to_kv_pairs(hash))
```
asatwal, line 83: If we have the `data` and `hidden_data` in in the original hash then we would need just the with_data methods as follows:

``` Ruby
def with_data(hash)
  @event_hash.deep_merge!(data: hash_to_kv_pairs(hash[:data]))
  @event_hash.deep_merge!(hidden_data: hash_to_kv_pairs(hash[:hidden_data]))
  self
end
```

asatwal (commented):
@ericaporter - I've suggested a slightly different approach in order to minimise changes, but I might be over simplifying here or misunderstood something. We can discuss anyway. Thanks

lib/dfe/analytics/initialisation_events.rb
```diff
@@ -27,7 +27,7 @@ def send_initialisation_events
 
         initialise_analytics_event = DfE::Analytics::Event.new
                                                 .with_type('initialise_analytics')
-                                                .with_data(initialise_analytics_data)
+                                                .with_data(data: initialise_analytics_data)
                                                 .as_json
```
ericaporter, line 31: This was necessary - tests failing without specifying `data:`

spec/dfe/analytics/entities_spec.rb
```diff
@@ -1,14 +1,16 @@
 # frozen_string_literal: true
 
 RSpec.describe DfE::Analytics::Entities do
-  let(:interesting_fields) { [] }
+  let(:allowlist_fields) { [] }
   let(:pii_fields) { [] }
```
ericaporter, line 5: Just took the opportunity to rename this, I find it a bit confusing

spec/dfe/analytics/load_entity_batch_spec.rb
```diff
@@ -36,7 +37,7 @@
       perform_enqueued_jobs do
         c = Candidate.create(email_address: '12345678910')
         c2 = Candidate.create(email_address: '12345678910')
-        stub_const('DfE::Analytics::LoadEntityBatch::BQ_BATCH_MAX_BYTES', 250)
+        stub_const('DfE::Analytics::LoadEntityBatch::BQ_BATCH_MAX_BYTES', 300)
```
ericaporter, line 41: Tests failing without this increase

spec/dfe/analytics/load_entity_batch_spec.rb
```diff
@@ -47,7 +48,7 @@
     it 'doesn’t split a batch unless it has to' do
       c = Candidate.create(email_address: '12345678910')
       c2 = Candidate.create(email_address: '12345678910')
-      stub_const('DfE::Analytics::LoadEntityBatch::BQ_BATCH_MAX_BYTES', 500)
+      stub_const('DfE::Analytics::LoadEntityBatch::BQ_BATCH_MAX_BYTES', 550)
```
ericaporter, line 52: Tests failing without this increase

lib/dfe/analytics/event.rb
```diff
@@ -73,7 +73,8 @@ def with_entity_table_name(table_name)
       end
 
       def with_data(hash)
-        @event_hash.deep_merge!(data: hash_to_kv_pairs(hash))
+        @event_hash.deep_merge!(data: hash_to_kv_pairs(hash[:data]))
+        @event_hash.deep_merge!(hidden_data: hash_to_kv_pairs(hash[:hidden_data]))
```
ericaporter, line 78: I guess the only downside to doing it this way is we now have a more complex data structure that might be a bit less performant

lib/dfe/analytics/event.rb
```diff
@@ -73,7 +73,8 @@ def with_entity_table_name(table_name)
...
```
asatwal, line 77: I think we might need to  only add `data` or `hidden_data` if it is present, as the initialise and checksum events don't add `hidden_data`
```suggestion
        @event_hash.deep_merge!(data: hash_to_kv_pairs(hash[:data])) if hash[:data].any?
        @event_hash.deep_merge!(hidden_data: hash_to_kv_pairs(hash[:hidden_data]))  if hash[:hidden_data].any?
```

lib/dfe/analytics/event.rb
ericaporter: Now updated

asatwal (commented):
Looks good @ericaporter - I just have a few questions about on the specs.

spec/dfe/analytics/entities_spec.rb
```diff
@@ -143,8 +165,7 @@
             'entity_table_name' => Candidate.table_name,
             'event_type' => 'update_entity',
             'data' => [
-              { 'key' => 'email_address', 'value' => ['bar@baz.com'] },
-              { 'key' => 'first_name', 'value' => ['Jason'] }
+              { 'key' => 'email_address', 'value' => ['bar@baz.com'] }
```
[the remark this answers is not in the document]
ericaporter: So previously we were but did that functionality make sense? We only update the email address so shouldn't we just get the updated email address? (Unless there is a benefit to sending the unchanged fields?)

spec/dfe/analytics/entities_spec.rb
```diff
@@ -203,5 +251,22 @@
...
+      let(:hidden_pii_fields) { %w[dob] }
+
+      it 'sends event indicating deletion with allowed and hidden data' do
+        entity = Candidate.create(email_address: 'to@be.deleted', dob: '21062000')
+        entity.destroy
+
+        expect(DfE::Analytics::SendEvents).to have_received(:perform_later)
+          .with([a_hash_including({
+            'event_type' => 'delete_entity',
+            'data' => array_including(a_hash_including('key' => 'email_address')),
```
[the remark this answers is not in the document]
ericaporter: Now updated

spec/dfe/analytics/entities_spec.rb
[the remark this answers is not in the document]
ericaporter: Checked with Steven and it is essential to send complete record of allowable fields... so reverted to previous logic with a few updates to method names to try to make it more clear

asatwal (approved):
👍 - Nice work.

PR #129, opened 2024-04-09
Add validation for hidden pii fields appearing on both lists

ericaporter:
[Ticket ](https://trello.com/c/UOsEQQSu/1878-prevent-fields-from-appearing-in-both-analyticshiddenpiiyml-and-analyticspiiyml)This PR adds a validation to prevent fields appearing in analytics_hidden_pii.yml and analytics_pii.yml

N.B. I'll change the base branch when separate-hidden-data has been merged into the epic branch

### DIFF
--- a/lib/dfe/analytics.rb
+++ b/lib/dfe/analytics.rb
@@ -143,8 +143,7 @@ module DfE
 
     def self.hidden_pii
       Rails.application.config_for(:analytics_hidden_pii)
-    rescue StandardError
-      Rails.logger.info('Missing configuration: analytics_hidden_pii.yml. Please create this file in the config directory')
+    rescue RuntimeError
       { 'shared' => {} }
     end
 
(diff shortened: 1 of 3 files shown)


asatwal (approved):
LGTM 👍

lib/dfe/analytics.rb
```diff
@@ -143,6 +143,9 @@ def self.allowlist_pii
 
     def self.hidden_pii
       Rails.application.config_for(:analytics_hidden_pii)
+    rescue StandardError
```
asatwal, line 146: We normally catch `RuntimeError` for these config errors, as it's a bit less catch all.

lib/dfe/analytics.rb
```diff
@@ -143,6 +143,9 @@ def self.allowlist_pii
...
+      Rails.logger.info('Missing configuration: analytics_hidden_pii.yml. Please create this file in the config directory')
```
asatwal, line 147: Could we possibly leave out the logging ? If a project doesn't have any hidden pii, then arguably we don't need a config file.

asatwal (approved):
👍 LGTM

PR #130, opened 2024-04-12
Add flags to prevent code running in envs where dfe-analytics isn't set up

ericaporter:
Events are being triggered on Register Sandbox env which isn't set up to send events to BigQuery. This is causing issues with lots of erroring jobs in Sentry, lots of emails, noise etc.

Adding in these flags should prevent events from sending in both the bulk upload and the entity table check job unless the env is set up.


spec/dfe/analytics/entity_table_check_job_spec.rb
```diff
@@ -20,9 +20,25 @@
...
+        described_class.new.perform
+      end
+    end
+
     context 'when entity table checks are not enabled' do
       before do
+        allow(DfE::Analytics).to receive(:enabled?).and_return(true)
         allow(DfE::Analytics).to receive(:entity_table_checks_enabled?).and_return(false)
+        allow(DfE::Analytics).to receive(:entities_for_analytics).and_return(%w[Candidate Application])
       end
```
ericaporter, line 42: I realised this test would always be false if the other allows were not added

lib/dfe/analytics/services/entity_table_checks.rb
```diff
@@ -94,6 +94,8 @@ def entity_table_check_data(entity_name, order_column)
         end
 
         def send_entity_table_check_event(entity_name, entity_type, entity_tag, order_column)
+          return unless DfE::Analytics.enabled?
+
```
ericaporter, line 98: Arguably this one isn't needed if we have a return in the entity_table_check_job

asatwal: This services gets called directly from the import rake tasks. I think this is still required, but is better moved to the call method. Or maybe we don't need at all if we put the guard in `import_entities.rake`

lib/dfe/analytics/load_entity_batch.rb
```diff
@@ -6,6 +6,11 @@ class LoadEntityBatch < AnalyticsJob
       BQ_BATCH_MAX_BYTES = 10_000_000
 
       def perform(model_class_arg, ids, entity_tag)
+        unless DfE::Analytics.enabled?
```
asatwal, line 9: Maybe we don't need at all if we put the guard in import_entities.rake - It only gets called from the rake task right ?

asatwal (commented):
Looks good. Just a few comments for you perusal

lib/dfe/analytics/tasks/import_entities.rake
```diff
@@ -2,6 +2,8 @@ namespace :dfe do
   namespace :analytics do
     desc 'Send Analytics events for the (allowlisted) state of all records in the database'
     task :import_all_entities, %i[batch_size] => :environment do |_, args|
+      return unless DfE::Analytics.enabled?
```
asatwal, line 5: Nitpick - Might be worth having a `puts`here writing to the console, with a message like:
```
DfE Analytics is not enabled - Ignoring import_all_entities
```

lib/dfe/analytics/tasks/import_entities.rake
```diff
@@ -11,6 +13,8 @@ namespace :dfe do
 
     desc 'Send Analytics events for the state of all records in a specified model'
     task :import_entity, %i[entity_name batch_size] => :environment do |_, args|
+      return unless DfE::Analytics.enabled?
```
asatwal, line 16: Nitpick - Might be worth having a `puts` here writing to the console, with a message like:
```
DfE Analytics is not enabled - Ignoring import_entity
```

asatwal (approved):
Looks good - Just a few optional messages to the console would be informative I think when the import rake tasks are called but analytics disabled.

lib/dfe/analytics/tasks/import_entities.rake
ericaporter: Added this 👍

lib/dfe/analytics/tasks/import_entities.rake
ericaporter: Added this 👍

PR #131, opened 2024-04-19
Add ImportEntityTableCheck to import task

ericaporter:
Previously we updated the bulk import task to send an EntityTableCheck event.

This PR adds that functionality to the single table import task.


asatwal (approved):
👍 LGTM

PR #132, opened 2024-04-19
Explicitly truncate updated_at and created_at

ericaporter:
We need to explicitly truncate updated_at and created_at timestamps to milliseconds to be able to compare checksums between production db tables and Dataform


asatwal (commented):
@stevenleggdfe @ericaporter 

The where clause here: 

https://github.com/DFE-Digital/dfe-analytics/blob/80015465040513020d0c2b3a2ae45d4f05f3b547/lib/dfe/analytics/shared/checksum_query_components.rb#L5

Which is used to filter out anything before checksum_calculated_at still uses microseconds precision for the order column. This will be different in BQ as it doesn't have microseconds. We just need to make this consistent also between dfe-analytics and BQ.

stevenleggdfe:
> @stevenleggdfe @ericaporter
> 
> The where clause here:
> 
> https://github.com/DFE-Digital/dfe-analytics/blob/80015465040513020d0c2b3a2ae45d4f05f3b547/lib/dfe/analytics/shared/checksum_query_components.rb#L5
> 
> Which is used to filter out anything before checksum_calculated_at still uses microseconds precision for the order column. This will be different in BQ as it doesn't have microseconds. We just need to make this consistent also between dfe-analytics and BQ.

Good point. It's easy enough for me to change that in dfe-analytics-dataform if @ericaporter you are happy to change in dfe-analytics as well?

asatwal (approved):
👍 LGTM

PR #134, opened 2024-04-22
Fix for flakey tests

ericaporter:
We are getting flakey tests in this file with the checksum_calculated_at being 1s after the expectation. This PR 
- Updates `@checksum_calculated_at` to store a Time object instead of a 
  string to ensure compatibility with Timecop.freeze
- Modifies the before block to fetch current_timestamp from the database, 
  convert it to the specified timezone, and directly use this Time object 
  with Timecop.freeze
- Ensure Timecop is used effectively by freezing time at the exact 
  moment defined by the database's current timestamp in the specified 
  timezone


spec/dfe/analytics/services/entity_table_checks_spec.rb
```diff
@@ -68,7 +67,11 @@
     let(:department_entity) { DfE::Analytics.entities_for_analytics.find { |entity| entity.to_s.include?('department') } }
     let(:entity_type) { 'entity_table_check' }
```
asatwal, line 68: Use instance variables in tests is normally discouraged. I would add a let for the checksum_calculated_at instead of using an instance variable in the tests.
```
  let(:checksum_calculated_at) { @checksum_calculated_at }
```

Then just use  checksum_calculated_at rather than @checksum_calculated_at

asatwal (approved):
LGTM - Minor comment regarding use of instance variables in tests.

PR #146, opened 2024-05-24
Cover edge cases for validations

ericaporter:
When testing the validation that hidden_pii fields do not overlap analytics_pii fields I noticed that the validation method would error if either the allowlist_pii file were in any of these scenarios:

1. `shared:`
2. `shared: candidates:`
3. `shared: candidates: []`

rather than `shared: {}` which is explicitly empty. To avoid this I updated the following methods to:
```
def self.hidden_pii
         DfE::Analytics.hidden_pii || {}
  end

def self.allowlist_pii
      DfE::Analytics.allowlist_pii || {}
end
```

and added in `next if fields.nil? || fields.empty?` to the overlap validation method

### DIFF
--- a/lib/dfe/analytics/fields.rb
+++ b/lib/dfe/analytics/fields.rb
@@ -78,11 +78,11 @@ module DfE
       def self.overlapping_pii_fields
         overlapping_fields = []
         hidden_pii.each do |entity, fields|
-          next if fields.nil? || fields.empty?
+          next if fields.blank?
 
           if allowlist_pii[entity]
             overlapping = fields & allowlist_pii[entity]
-            overlapping_fields.concat(overlapping) unless overlapping.empty?
+            overlapping_fields.concat(overlapping) unless overlapping.blank?
           end
         end
         overlapping_fields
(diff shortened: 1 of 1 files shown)


lib/dfe/analytics/fields.rb
```diff
@@ -68,16 +68,18 @@ def self.allowlist
...
 
       def self.allowlist_pii
-        DfE::Analytics.allowlist_pii
+        DfE::Analytics.allowlist_pii || {}
       end
 
       def self.overlapping_pii_fields
         overlapping_fields = []
         hidden_pii.each do |entity, fields|
+          next if fields.nil? || fields.empty?
```
asatwal, line 81: Nitpick:
I believe `fields.blank` is the same as `fields.nil? || fields.empty?`

asatwal (approved):
LGTM 👍 - Just one minor nitpick.

lib/dfe/analytics/fields.rb
ericaporter: Yes! Now updated! Can update `unless overlapping.empty?`to unless overlapping.blank? too since overlapping will always be an array

PR #147, opened 2024-06-04
Fix import_all_entities rake task output

slorek:
When `enable_analytics` is `true` the `import_all_entities` rake task outputs a confusing message stating that it is not enabled, but nonetheless continues to perform the task:

```bash
$ bundle exec rails dfe:analytics:import_all_entities
DfE Analytics is not enabled - Ignoring import_all_entities
```

This fix should avoid some confusion for users! :)


ericaporter (approved):
LGTM 👍

PR #150, opened 2024-06-17
Update log masking method

ericaporter:
This PR includes updates to the masking hidden_pii strategy

1. It removes masking_hidden_data method from the event_matcher as this isn't necessary for masking the logs
2. Updates the masking method, so instead of finding the hidden_pii fields per entity, it now just searches for the hidden_pii key/values in the logs
3. Updates to filter_matched? method to avoid exceptions


asatwal (approved):
👍 LGTM

PR #151, opened 2024-06-24
Add big_query_api requires to testing.rb

ericaporter:
Schools experience were seeing an error when running the test suite after upgrading to the latest version of dfe-analytics.

This addition should prevent other services from seeing the same error.


asatwal (approved):
👍

PR #158, opened 2024-07-29
Update callbacks to have _commit suffix

ericaporter:
ECF reported that they were seeing a disparity between production db numbers and BigQuery arising during a transaction - the event is sending straight away so if the transaction ends up rolling back the analytics event has already been sent to BQ and doesn’t match the db

This PR updates the callbacks (after_create after_update and after_destroy) to be after_create_commit `after_update_commit and after_destroy_commit so that events will only send after the transaction is complete but also if it rolls back the event will not send`

ECF did some testing of this update:

The issue was initially detected in the method save_and_dedupe_participant https://github.com/DFE-Digital/early-careers-framework/blob/main/app/models/npq_application.rb#L134. I've tested the fix locally by using some binding.pry statements to highlight where the analytics log events are occurring and if it's where we expect.
Pic 1: Before the change. Here we can see the event is logged before the transaction is complete. We've hit the pry while still in the transaction but the event has already been sent.
Pic 2: Now with the gem update. We hit the first pry we had before inside the transaction without logging an analytics event and we then publish the event before hitting the second pry outside the transaction. This is what we want. After the first, before the second.
Pic 3: Also with the gem update. I raised a rollback error inside the transaction and we completed the transaction without sending any event whatsoever which again is what we want.
Screenshot 2024-07-29 at 16 52 30
Screenshot 2024-07-29 at 16 45 14
Screenshot 2024-07-29 at 16 32 46

### DIFF
--- a/lib/dfe/analytics/entities.rb
+++ b/lib/dfe/analytics/entities.rb
@@ -8,20 +8,17 @@ module DfE
       included do
         attr_accessor :event_tags
 
-        after_create do
+        after_create_commit do
           extracted_attributes = DfE::Analytics.extract_model_attributes(self)
           send_event('create_entity', extracted_attributes) if extracted_attributes.any?
         end
 
-        after_destroy do
+        after_destroy_commit do
           extracted_attributes = DfE::Analytics.extract_model_attributes(self)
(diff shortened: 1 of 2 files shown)


asatwal (approved):
👍 Nice work - LGTM.

PR #162, opened 2024-08-15
Resolve undefined method `migration_context` when upgrading to rails 7.2

avinhurry:
### Context

We are getting the following error when upgrading our apps to rails version 7.2:

```
NoMethodError: undefined method `migration_context' for #<ActiveRecord::ConnectionAdapters::PostgreSQLAdapter:0x0000000000c620

raise ActiveRecord::PendingMigrationError if ActiveRecord::Base.connection.migration_context.needs_migration?

```

- Publish: https://github.com/DFE-Digital/publish-teacher-training/pull/4445

- Apply: https://github.com/DFE-Digital/apply-for-teacher-training/pull/9692

- Register: https://github.com/DFE-Digital/register-trainee-teachers/pull/4562

This is because of breaking changes and deprecation in rails 7.2:

- **[The `connection` method was changed to `with_connection`](https://blog.saeloun.com/2024/07/01/rails-adds-activerecord-base-with-connection)**

- **[The oldest Ruby version required for rails 7.2 is 3.1](https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html )**

- **[The deprecation of `fixture_path` meant that rspec had to be upgraded](https://github.com/rspec/rspec-rails/issues/2661)**

**Dropping Ruby 2.7 tests in CI** as that version has [reached it's EOL](https://www.ruby-lang.org/en/downloads/branches/).

image

**This PR also sets functional versions of the `rspec-rails` and `rails gems` depending on what version of rails you are using.**


asatwal (approved):
👍  LGTM

ericaporter (approved):
LGTM 👍  thanks for picking this up @avinhurry

PR #164, opened 2024-09-05
Allow ordering by id if created_at/updated_at is null

ericaporter:
[Ticket](https://trello.com/c/F6HB3sf3/2099-investigate-rowcount-mismatches-on-findpub-and-teaching-vacancies)
Publish have a table "course_subjects" that has been long term out of sync with Big Query
Investigation showed this was because they have a number of records where updated_at is null and they were being excluded from the row_count during entity table check job / import entity rake task causing the row_count and checksum to not match
This is thought to apply to several other tables in Publish and to at least one other service

This PR updates:

 1) the entity_table_check_job to firstly order by created_at if updated_at has null values, and finally to default to ID if updated_at/created_at are missing or have null values
 2) the where clause to include order_by records with null values so that we get an accurate row count

### DIFF
--- a/lib/dfe/analytics/services/entity_table_checks.rb
+++ b/lib/dfe/analytics/services/entity_table_checks.rb
@@ -67,15 +67,23 @@ module DfE
         end
 
         def determine_order_column(entity_name, columns)
-          if connection.column_exists?(entity_name, :updated_at) && columns.include?('updated_at')
+          if connection.column_exists?(entity_name, :updated_at) && columns.include?('updated_at') && !null_values_in_column?('updated_at')
             'UPDATED_AT'
-          elsif connection.column_exists?(entity_name, :created_at) && columns.include?('created_at')
+          elsif connection.column_exists?(entity_name, :created_at) && columns.include?('created_at') && !null_values_in_column?('created_at')
             'CREATED_AT'
           elsif connection.column_exists?(entity_name, :id) && columns.include?('id')
             'ID'
           else
             Rails.logger.info("DfE::Analytics: Entity checksum: Order column missing in #{entity_name}")
           end
+        end
+
+        def null_values_in_column?(column)
+          connection.select_value(<<-SQL).to_i.positive?
+            SELECT COUNT(*)
+            FROM #{connection.quote_table_name(entity_name)}
+            WHERE #{column} IS NULL
+          SQL
         end
 
         def entity_table_check_data(entity_name, order_column)
--- a/lib/dfe/analytics/services/postgres_checksum_calculator.rb
+++ b/lib/dfe/analytics/services/postgres_checksum_calculator.rb
@@ -18,7 +18,9 @@ module DfE
         end
 
         def call
-          calculate_checksum
+          ActiveRecord::Base.transaction do
+            calculate_checksum
+          end
         end
 
         private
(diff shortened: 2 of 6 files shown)


lib/dfe/analytics/services/postgres_checksum_calculator.rb
```diff
@@ -65,7 +67,8 @@ def build_select_and_order_clause(order_column, table_name_sanitized)
         def build_where_clause(order_column, table_name_sanitized, checksum_calculated_at_sanitized)
           return '' unless WHERE_CLAUSE_ORDER_COLUMNS.map(&:downcase).include?(order_column.downcase)
 
-          "WHERE DATE_TRUNC('milliseconds', #{table_name_sanitized}.#{order_column.downcase}) < DATE_TRUNC('milliseconds', #{checksum_calculated_at_sanitized}::timestamp)"
+          # Add IS NULL to include records with null updated_at / created_at values
+          "WHERE (#{table_name_sanitized}.#{order_column.downcase} IS NULL OR DATE_TRUNC('milliseconds', #{table_name_sanitized}.#{order_column.downcase}) < DATE_TRUNC('milliseconds', #{checksum_calculated_at_sanitized}::timestamp))"
```
stevenleggdfe, line 71: @ericaporter will this addition ever actually do anything, since if it is true then the WHERE clause wouldn't operate, or would be switched to created_at if created_at does not contain null values?

lib/dfe/analytics/services/entity_table_checks.rb
```diff
@@ -78,6 +78,14 @@ def determine_order_column(entity_name, columns)
           end
         end
 
+        def null_values_in_column?(column)
+          connection.select_value(<<-SQL).to_i.positive?
+            SELECT COUNT(*)
```
stevenleggdfe, line 83: @ericaporter maybe check whether being specific as to any field name instead of using * in the COUNT() is more efficient? Certainly in column stores avoiding COUNT(*) is a huge efficiency boost but this is running in Postgres which is a row store.

e.g. ```SELECT COUNT(#{column})```...

lib/dfe/analytics/services/postgres_checksum_calculator.rb
ericaporter: I think that's fair, this was the first thing I updated but it won't do anything if we are excluding order columns with null values, as much as I like a belt and braces approach...

lib/dfe/analytics/services/entity_table_checks.rb
ericaporter: @stevenleggdfe I opted for COUNT(*) because I thought COUNT{column} would ignore null values...

stevenleggdfe: So it does! That is one scary difference between PostgreSQL and GoogleSQL. Ignore me :-)

lib/dfe/analytics/services/postgres_checksum_calculator.rb
asatwal: OK to keep as belt and braces I believe.

asatwal (commented):
Looks good @ericaporter - Just a few comments.

lib/dfe/analytics/services/entity_table_checks.rb
ericaporter: @asatwal @stevenleggdfe we do another COUNT(*) in the main checksum calculation and I considered having both as count(id) but I couldn't work out of there was a possibility we might have null id values that we should be including. If the performance benefits would outweigh this likelihood, I'm happy to update, but as this fix is about not excluding nulls, I felt it was tempting fate a little 😄

asatwal (approved):
👍

PR #166, opened 2024-09-19
Remove ability to pseudonymise data

ericaporter:
We no longer wish to support the ability to pseudonymise data in the gem since we now send it separately as hidden_pii.

This PR removes reference to pseudonymisation from the docs and the codebase.

**N.B. I have left in the anonymise method in analytics.rb since it is being used by anonymised_user** _agent_and_ip: anonymised_user_agent_and_ip(rack_request) which I have not removed at this stage.**

**N.B. **For deployment,** first check whether any service is still sending pseudonymised data... we would ideally have them sending that data as hidden_pii prior to deployment. Some general comms to services might be helpful


docs/create-events-table.sql
```diff
@@ -13,7 +13,7 @@ CREATE TABLE
...
+    value ARRAY < STRING > OPTIONS(description="Contents of the field in the database after it was created or updated, or just before it was imported or destroyed.") > > OPTIONS(description="ARRAY of STRUCTs, each with a key and a value. Contains a set of data points appropriate to the event_type of this event. For example, if this event was an entity create, update, delete or import event, data will contain the values of each field in the database after this event took place - according to the settings in the analytics.yml configured for this instance of dfe-analytics."),
```
stevenleggdfe, line 16: Good spot!

lib/dfe/analytics/fields.rb
```diff
@@ -71,23 +62,6 @@ def self.hidden_pii
         DfE::Analytics.hidden_pii || {}
       end
 
-      def self.allowlist_pii
```
stevenleggdfe, line 74: @ericaporter do we still have the concept of requiring each field in each table to be in *either* analytics_blocklist.yml or analytics.yml, plus optionally analytics_hidden_pii.yml? Enforcement of that first 'or' needs to remain.

I think you probably haven't changed that here but wanted to double check!

ericaporter: We've got the originals validations still which are:
1) fields appearing in analytics.yml that have been removed
2) new fields detected that aren't in analytics.yml or blocklist.yml
3) fields existing in both analytics & blocklist

lib/dfe/analytics/event.rb
```diff
@@ -126,15 +126,15 @@ def hash_to_kv_pairs(hash)
       end
 
       def anonymised_user_agent_and_ip(rack_request)
-        DfE::Analytics.pseudonymise(rack_request.user_agent.to_s + rack_request.remote_ip.to_s) if rack_request.remote_ip.present?
+        DfE::Analytics.anonymise(rack_request.user_agent.to_s + rack_request.remote_ip.to_s) if rack_request.remote_ip.present?
```
ericaporter, line 129: @asatwal I left in this method but reverted the naming to anonymise. This was there prior to your addition of pseudonymise_web_request_user_id, although that's when the naming changed to pseudonymise. I can remove this too and the associated tests but kept it in for time being as it seemed useful if is actually being used

stevenleggdfe: @ericaporter I'd suggest we call it something that is aligned with the algorithm. If it's doing a one way hash I'd say pseudonymise; if it's overwriting that part of the string with (say) REDACTED then I'd go anonymise.

This is significant because pseudonymised data stays PII under GDPR; anonymised data doesn't.

asatwal: I would say anonymise is OK as this user agent + ip is not really PII.

asatwal (approved):
👍 LGTM

PR #169, opened 2024-11-06
Config option to exclude models

slawosz:
This pull request introduces a config option to disable one or more models entirely:

```
proc { |x| x.to_s =~ /Namespace::/ } will exclude all
```

In my application, there were 2 models with very similar names: `Namespace::Foo` and `Foo`, pointing to table `foos` in 2 different databases. Because `dfe-analytics` does not recognize namespaces or different databases, the above option is a viable solution to the problem. The reason for similar models/tables is migration - one application data is moving to another.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -98,3 +98,10 @@ en:
           description: |
             Google generated cloud credentials file
           default: ENV['GOOGLE_CLOUD_CREDENTIALS']
+        excluded_models_proc:
+          description: |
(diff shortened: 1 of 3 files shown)


asatwal:
👍 LGTM. Looks like a useful change also. Some rspec failures though.

asatwal (commented):
👍 LGTM. Looks like a useful change also. Some rspec failures though.

asatwal (approved):
👍

PR #175, opened 2024-11-20
headers['X-REAL-IP']

goodviber:
We need to change the remote_ip which is dynamic to the X-REAL-IP header which is static over a session. 

Should this be HTTP_X_REAL_IP?

This will need in app testing, via a review app?


asatwal (commented):
Looks good. 

Maybe need to check with DevOps (@saliceti) that the X-REAL-IP is available in the header for all TS Systems that use DfE::Analytics.

goodviber:
Hi Amar, I spoke with Colin Saliceti prior to starting the ip ticket, the X-REAL-IP is an Azure specific ip, so should be present in all services as long as we are on Azure, if we change platforms we could change to the first element of X-Forwarded-For header list as that is agnostic. Colin favoured the X-REAL-IP.

Nitemaeric (approved):

asatwal (approved):
👍 LGTM

PR #181, opened 2024-12-04
Fix "async" config setup

scruti:
`DfE::Analytics` initialisation code ignored the `config.async = false` option when provided.

This was caused by the conditional assignment operator `||=`, as `foo ||= true` does set `foo = true` if it was either `nil` or `false`.

Changed it to default to `true` only if no explicit value was set.


scruti:
Tests pass locally (at least on my setup!)
image

PR #186, opened 2024-12-16
update transaction tests

goodviber:
The transaction rollback tests seem to have been incorrectly configured, calling a create action outside of the transaction block, which results in a send event even when the block is rolled back.


asatwal (approved):
👍 LGTM

PR #188, opened 2024-12-18
Fix after_update_commit

goodviber:
We have noticed that after_update_commits were not passing along any data to BQ. This update fixes this issue by introducing a custom TransactionChanges module which adds to  the default rails behaviour. 

We can now call `transaction_changed_attributes`, in additon to `saved_changes, previous_changes` etc which did not suit our purpose in transactions `after_update_commit`


lib/dfe/analytics/transaction_changes.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
```
asatwal, line 1: I'm not sure I follow this entirely as it's a bit esoteric. Is this an existing solution to the problem ?

spec/dfe/analytics/transaction_changes_spec.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
```
asatwal, line 1: Do we need to test updates to multiple tables also ?

asatwal (commented):
Just a few comments below. My main concern here is the rollout of this fix. If we could build that into DfE::Analytics without a change to `dfe_analytics.rb` so that BAU release could fix the problem.

README.md
```diff
@@ -224,6 +224,23 @@ Once `analytics.yml` and friends are populated, the associated models will be
 automatically instrumented to send updates to BigQuery via ActiveRecord
 callbacks.
 
+#### Transactions
+
+In order to correctly record data changes in transactions, specifically in regard to rollbacks and updates, we need to include the TransactionChanges module in all ActiveRecord models, so ensure this is present in config/initializers/dfe-analytics.rb
+
+```ruby
+ActiveSupport.on_load(:active_record) do
+  include DfE::Analytics::TransactionChanges
```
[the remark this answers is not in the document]
goodviber: Good idea, I will spin it up

lib/dfe/analytics/transaction_changes.rb
goodviber: Yes its mainly from https://github.com/dylanahsmith/ar_transaction_changes/blob/main/README.md

spec/dfe/analytics/transaction_changes_spec.rb
goodviber: Good idea, I will spin it up

README.md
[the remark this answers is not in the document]
goodviber: I have put it in the initialize! , the railtie didn't seem to work correctly

spec/dfe/analytics/transaction_changes_spec.rb
goodviber: It is a bit difficult to set up 2 tables in the test, as we use the before block to initialise with the allowed/blocklist etc. However there is no reason 2 tables wouldn't update in a transaction, and we can see this happening when we log the output of the SubmitApplicationChoice transaction in [Apply](https://github.com/DFE-Digital/apply-for-teacher-training/blob/75213a81621c929ef79e659948766033a04f9139/app/services/candidate_interface/submit_application_choice.rb#L15), we get:
```
2025-01-15 10:42:13.769163 I [67740:puma srv tp 003] Rails -- log_updated_attributes: {:data=>{"submitted_at"=>2025-01-15 10:42:13.509937000 GMT +00:00, "updated_at"=>2025-01-15 10:42:13.653660000 GMT +00:00}}
2025-01-15 10:42:13.769181 I [67740:puma srv tp 003] Rails -- log_model: #<ApplicationForm:0x00000001342b4398>
2025-01-15 10:42:13.775288 I [67740:puma srv tp 003] Rails -- log_updated_attributes: {:data=>{"reject_by_default_at"=>2025-02-26 23:59:59.999999999 GMT +00:00, "reject_by_default_days"=>30, "sent_to_provider_at"=>2025-01-15 10:42:13.551772000 GMT +00:00, "status"=>"awaiting_provider_decision", "updated_at"=>2025-01-15 10:42:13.646503000 GMT +00:00}}
2025-01-15 10:42:13.775300 I [67740:puma srv tp 003] Rails -- log_model: #<ApplicationChoice:0x00000001342b1f58>
```

lib/dfe/analytics/transaction_changes.rb
goodviber: It basically adds the transaction_changed_attributes to a model

README.md
```diff
@@ -224,6 +224,17 @@ Once `analytics.yml` and friends are populated, the associated models will be
...
 
+#### Transactions
+
+In order to correctly record data changes in transactions, specifically in regard to rollbacks and updates, we include the TransactionChanges module in all ActiveRecord models.
+
+This can conflict with other gems who modify the `active_record#write_attribute` method, e.g. `globalize` gem. The solution is to require this gem first, e.g.
+
+```
+gem 'dfe-analytics'
+gem 'globalize' require: ['dfe-analytics', 'globalize']
```
asatwal, line 235: Do we know if this is an existing problem on any project (Which we probably don't!)? If it is then what will be implications if we don't fix it ? My feeling is that the issue we're trying to fix is not fixed and goes unnoticed.

asatwal (commented):
Have these changes been tested in situ and is existing functionality OK ? ie No regression. I think this has been done in Apply right ?

goodviber:
I have been running it against a local version of Apply, and all rspec tests pass. When interacting with the app, I have only been focussing on the application form process.

README.md
goodviber: I don't have much insight on this, globalize is not the only gem which may define write_attribute. I would hope we would see some failing tests in the build if there was a conflict?

asatwal (approved):
Looks good. As I mentioned, I believe it would be good to ensure that there is no regression in a real environment. Register is always a good candidate.

PR #192, opened 2025-04-03
update error levels

goodviber:
We are rescuing initialisation and other errors with Rails.logger.info messages, which makes them difficult to find In the output logs. This PR updates the relevant info messages to error or warn level


asatwal (approved):
👍  LGTM

PR #195, opened 2025-04-10
Make the gem work without ActiveRecord

hrmtl:
The `dfe-analytics` gem currently assumes the presence of a database and `ActiveRecord` in any app that uses it.

This PR adds a few modifications to allow the gem to work without a database.

The motivation for this came when we started to incorporate the gem into the [GHBS FABS Rails app](https://github.com/DFE-Digital/find-a-buying-solution). Our app doesn't use a database. It uses Contentful CMS as its backend.

We only need the web requests tracking feature of `dfe-analytics`.

This PR makes the following modifications to allow the use of the gem without a database:

* Check if ActiveRecord is defined before loading and running any dependent code.
* Group together files which use ActiveRecord.
* Update the initializer log message.
* Add a test

As per my discussion with @asatwal, there may be scope to do a bigger refactor to modularise all the AR-related code. However, I did not want to make any big changes with my first PR seeing as many projects are using this gem.

This PR does just enough to make the gem usable without AR. Any feedback or suggestions are welcome.


Gemfile.lock
```diff
@@ -244,17 +245,24 @@ GEM
     net-smtp (0.5.0)
       net-protocol
     nio4r (2.7.4)
-    nokogiri (1.16.7-aarch64-linux)
+    nokogiri (1.18.7)
```
asatwal, line 248: Not sure why nokogiri is being upgraded here, might be to do with the local config.

Gemfile.lock
```diff
@@ -399,16 +407,16 @@ GEM
...
-    sqlite3 (2.2.0-aarch64-linux-musl)
-    sqlite3 (2.2.0-arm-linux-gnu)
-    sqlite3 (2.2.0-arm-linux-musl)
-    sqlite3 (2.2.0-arm64-darwin)
-    sqlite3 (2.2.0-x86-linux-gnu)
-    sqlite3 (2.2.0-x86-linux-musl)
-    sqlite3 (2.2.0-x86_64-darwin)
-    sqlite3 (2.2.0-x86_64-linux-gnu)
-    sqlite3 (2.2.0-x86_64-linux-musl)
+    sqlite3 (2.6.0-aarch64-linux-gnu)
```
asatwal, line 410: Not sure why sqlite3 is being added here.

lib/dfe/analytics.rb
```diff
@@ -96,19 +92,21 @@ def self.configure
...
         message = "Warning: DfE Analytics is not set up. Run: 'bundle exec rails generate dfe:analytics:install'"
         Rails.logger.error(message)
         puts message
         return
       end
 
+      if defined?(ActiveRecord)
+        setup_entities
+      else
+        Rails.logger.info('ActiveRecord not loaded; DfE Analytics will only track web requests.')
```
asatwal, line 105: I think the message should say:
`DfE Analytics will only track non database events`
as we have custom events as well as web request events in some services.

lib/dfe/analytics.rb
```diff
@@ -96,19 +92,21 @@ def self.configure
...
-      end
-
       unless Rails.env.production? || File.exist?(Rails.root.join('config/initializers/dfe_analytics.rb'))
         message = "Warning: DfE Analytics is not set up. Run: 'bundle exec rails generate dfe:analytics:install'"
         Rails.logger.error(message)
         puts message
         return
       end
 
+      if defined?(ActiveRecord)
```
asatwal, line 102: We may want to introduce a config item, such as `send_db_events` to control this rather than just relying on ActiveRecord being present or not. TBD.

asatwal (commented):
👍 Looks like a good change to not rely on `ActiveRecord`.

@hrmtl - Have you tested that this works in an environment that does not have `ActiveRecord` ?

@goodviber - Could you review also please and what's your opinion on having a config item also to control sending db events ?

lib/dfe/analytics.rb
hrmtl: I did try that but the gem code would throw an error while loading in the Rails app. I agree that a config item would be nice, but maybe we can do it as an improvement in the future.

goodviber: What was the error?

hrmtl: I got a `NameError` for ActiveRecord when starting the Rails server. I think that happened when the gem code was loaded before the config got read.

hrmtl:
@asatwal Yes, I've tested it in the FABS Rails app which doesn't have ActiveRecord. I was able to successfully track web requests into BigQuery.

hrmtl:
@asatwal If you have CI setup here, can we run the test suite please?

lib/dfe/analytics.rb
goodviber: @asatwal I think we could look at the config issue later, I will have a tilt at it myself. If this has been tested @hrmtl with an app running ActiveRecord we can approve

asatwal (approved):
👍 - Nice work. Looks good to me.

hrmtl:
@asatwal Thanks!

PR #200, opened 2025-06-02
rm updated_at in checksum calculators

goodviber:
We are getting chacksum mismatched where we have identical row counts but differing updated at values for various reasons. We have decided to remove the updated_at values from the checksum calculators.

I think this will also need to be matched in dataform

### DIFF
--- a/spec/dfe/analytics/services/entity_table_checks_spec.rb
+++ b/spec/dfe/analytics/services/entity_table_checks_spec.rb
@@ -60,7 +60,7 @@ RSpec.describe DfE::Analytics::Services::EntityTableChecks do
 
   describe '#call' do
     let(:time_zone) { 'London' }
-    let(:order_column) { 'UPDATED_AT' }
+    let(:order_column) { 'CREATED_AT' }
     let(:course_entity) { DfE::Analytics.entities_for_analytics.find { |entity| entity.to_s.include?('course') } }
     let(:institution_entity) { DfE::Analytics.entities_for_analytics.find { |entity| entity.to_s.include?('institution') } }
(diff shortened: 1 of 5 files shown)


stevenleggdfe (commented):
Makes sense to me.

I wonder if there would be value in removing created_at ordering as well and just keeping it simple? i.e. ordering by id in all cases. We could then do away with order_column entirely. It would also substantially simplify things on the dfe-analytics-dataform end as well as reduce costs and query execution time.

I don't think ordering by created_at adds anything by itself unless as a proxy for updated_at if it happens to be missing.

spec/dfe/analytics/services/entity_table_checks_spec.rb
```diff
@@ -294,14 +296,13 @@
...
-      Candidate.create(email_address: 'first@example.com', updated_at: time_base + 0.001.seconds)
-      Candidate.create(email_address: 'second@example.com', updated_at: time_base + 0.005.seconds)
+    it 'orders records by created_at truncated to milliseconds' do
+      Candidate.create(email_address: 'first@example.com')
+      Candidate.create(email_address: 'second@example.com')
 
       described_class.call(entity_name: candidate_entity, entity_type: entity_type, entity_tag: nil)
 
-      ordered_candidates = Candidate.order(:updated_at).pluck(:email_address)
+      ordered_candidates = Candidate.order(:created_at).pluck(:email_address)
```
asatwal, line 305: This test may be flaky at times if the 2 records are created in the same millisecond. Not sure how you can fix this if passing in `created_at` is ignored.

asatwal (approved):
Looks good. Just one comment on possible increase in test flakyness (If that's a word :-))

PR #208, opened 2025-09-26
Change attributes

goodviber:
@jebw helpfully pointed out we do not need to overwrite the `changed_attributes` method in entities.rb, I have renamed this to `all_changed_attributes`

### DIFF
--- a/README.md
+++ b/README.md
@@ -552,7 +552,7 @@ Please note that page caching is project specific and each project must carefull
 
 ## Data health import issues
 
-1. Avoid the use of default scopes, used named scopes if possible. Otherwise the import task will not  catch unscoped entities
+1. Avoid the use of default scopes, used named scopes if possible. Otherwise the import task will not catch unscoped entities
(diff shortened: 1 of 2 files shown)


jebw (commented):
Since this is getting injected into the majority of data models across CPD / Teacher services I think you need to be quite specific to avoid risking naming conflicts - how about 

```ruby
def changed_attributes_for_dfe_analytics
  transaction_changed_attributes.each.with_object({}) do |attr, changes_with_values|
    changes_with_values[attr] = read_attribute(attr)
  end
end
```

Not sure if swapping to `read_attribute` is a good idea but would help avoid using `send` and presumably your trying to get a the db column values anyway?

jebw (approved):
LGTM

asatwal (commented):
@goodviber  Do we need any specific specs for this change to test the issue being fixed ?

goodviber:
@asatwal  No, but it seems we have to re-run the specs more than previously...

asatwal (approved):
LGTM 👍

PR #209, opened 2025-09-30
config added ignore_default_scope

goodviber:
we have added a config option to ignore default scopes when importing model instances to BQ


goodviber:
good point...i'll have a think

slorek:
In Claim we do want to send soft deleted records, so this works for us, however I wonder if this should be limited per-model rather than a global configuration. I don't think it will affect us on Claim, however.

goodviber:
@slorek I will have another think...

goodviber:
@slorek @asatwal have added a per model config as well the global all models option

asatwal (commented):
👍  LGTM

PR #233, opened 2026-05-06
v1.16.16

asatwal:
Release containing the following fixes:
- Bump actions/checkout from 2 to 
- Add dependabot yml 
- Restore config between runs
- Fixed spelling mistake in method call

### DIFF
--- a/lib/dfe/analytics/version.rb
+++ b/lib/dfe/analytics/version.rb
@@ -2,6 +2,6 @@
 
 module DfE
   module Analytics
-    VERSION = '1.15.15'
+    VERSION = '1.16.16'
   end
 end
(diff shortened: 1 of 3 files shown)


mysteryGarlicPress (approved):

lib/dfe/analytics/version.rb
```diff
@@ -2,6 +2,6 @@
 
 module DfE
   module Analytics
-    VERSION = '1.15.15'
```
Lockyy, line 5: Hi, just wanted to note that this version is listed as 1.16.16 in the repo and the changelog but the actual version tag is 1.15.16, I'm using the latter for pulling down this version but wanted to check which was the intentional version for future updates.

PR #237, opened 2026-09-01
Remove airbyte config json from gem config, it is now generated dynam…

mysteryGarlicPress:
…ically during airbyte rake tasks


lib/dfe/analytics/tasks/fields.rake
```diff
@@ -12,13 +12,5 @@ namespace :dfe do
         { shared: DfE::Analytics::Fields.generate_blocklist }.to_yaml
       )
     end
-
-    desc 'Generate a new airbyte stream config listing all fields to be send to Bigquery from analytics.yml'
-    task regenerate_airbyte_stream_config: :environment do
-      File.write(
-        DfE::Analytics.config.airbyte_stream_config_path,
-        DfE::Analytics::AirbyteStreamConfig.generate_pretty_json_for(DfE::Analytics.allowlist)
```
asatwal, line 20: I was thinking maybe we could leave a modified version of this task ie:
``` Ruby
puts DfE::Analytics::AirbyteStreamConfig.generate_pretty_json_for(DfE::Analytics.allowlist)
```

This allows us to generate the config for debug puposes just in case we need to do that. What do you think ?

lib/dfe/analytics/config.rb
asatwal, line 38: We can remove this also right ?

lib/dfe/analytics/config.rb
```diff
@@ -76,11 +76,10 @@ def self.configure(config)
         config.airbyte_server_url               ||= ENV.fetch('AIRBYTE_SERVER_URL', nil)
         config.airbyte_configuration            ||=
           JSON.parse(ENV.fetch('AIRBYTE_CONFIGURATION', '{}')).transform_keys(&:underscore).symbolize_keys
+        config.airbyte_stream_config_path       ||= nil
```
asatwal, line 79: We don't need this if we remove the config item `airbyte_stream_config_path` totally.

lib/dfe/analytics/fields.rb
```diff
@@ -45,20 +45,6 @@ def self.check!
           HEREDOC
         end
 
-        if DfE::Analytics.airbyte_enabled? && airbyte_conflicting_fields.any?
```
asatwal, line 48: We should be able to remove the method `airbyte_conflicting_fields` also right ?

asatwal (changes requested):
See comments in the review. Thanks

asatwal (approved):
Looks Good.

PR #239, opened 2026-09-07
Support multiple policy tags

asatwal:
Context
Support mutiple policy tables defined as JSON in ENV Variable.

Trello ticket: https://trello.com/c/BxYQWdrR

Changes proposed in this pull request:

- new format of `analytics_hidden_pii.yml file` is supported 
- Policy tags are loaded from `BIGQUERY_POLICY_TAG` ENV variable
- Backwards compatibility is maintained


mysteryGarlicPress (approved):
Happy to approve, however it would be good to have a safeguard against misspelled/missformed policy tag json. I presume:

- when json is wrong, the app wont start
- when policy tag does not exist, the update will fail

PR #240, opened 2026-09-08
Fix json version

mysteryGarlicPress:
json 3.0 removed the :quirks_mode option and raises on it; httparty's response parser still passes it (httparty/parser.rb).


asatwal (approved):
👍 LGTM

PR #241, opened 2026-09-10
Fix disable database events

asatwal:
Ensure database events are not sent which the configuration is set.

### DIFF
--- a/lib/dfe/analytics/config.rb
+++ b/lib/dfe/analytics/config.rb
@@ -68,7 +68,7 @@ module DfE
         config.azure_federated_auth             ||= false
         config.excluded_paths                   ||= []
         config.excluded_models_proc             ||= proc { |_model| false }
-        config.database_events_enabled          ||= true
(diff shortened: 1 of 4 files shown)


lib/dfe/analytics/config.rb
```diff
@@ -68,7 +68,7 @@ def self.configure(config)
         config.azure_federated_auth             ||= false
         config.excluded_paths                   ||= []
         config.excluded_models_proc             ||= proc { |_model| false }
-        config.database_events_enabled          ||= true
+        config.database_events_enabled            = true if config.database_events_enabled.nil?
```
mysteryGarlicPress, line 71: can do `config.database_events_enabled = config.database_event_enabled.nil?`

mysteryGarlicPress (approved):

lib/dfe/analytics/config.rb
asatwal: I find that a bit confusing, even though it's more compact.

